PR #2, opened 2019-04-26
Renamed INT to Int

emmberk:


emmberk:
merging without a build because this change won't affect the build

PR #7, opened 2019-05-02
Added Sonarcloud stage to the Jenkins file

jhunzik:
We can only run SonarQube on the master branch, so this needs to be merged to make sure it works. The build will kick off nightly as well as whenever someone pushes a change to master.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,9 +16,13 @@ pipeline {
         disableConcurrentBuilds()
         timestamps()
     }
+    triggers {
+        cron(env.BRANCH_NAME == "master" ? "H H(21-23) * * *" : "")
+    }
     environment {
         GITHUB_USERNAME = 'connexta'
         GITHUB_REPONAME = 'multi-int-store'
+        COVERAGE_EXCLUSIONS = '**/test/**/*,**/itests/**/*,**/*Test*,**/sdk/**/*,**/*.js,**/node_modules/**/*,**/jaxb/**/*,**/wsdl/**/*,**/nces/sws/**/*,**/*.adoc,**/*.txt,**/*.xml'
     }
     stages {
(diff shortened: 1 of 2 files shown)


jhunzik:
build now

Jenkinsfile
```diff
@@ -16,9 +16,13 @@ pipeline {
         disableConcurrentBuilds()
         timestamps()
     }
+    triggers {
+        cron(env.BRANCH_NAME == "master" ? "H H(21-23) * * *" : "")
+    }
     environment {
         GITHUB_USERNAME = 'connexta'
         GITHUB_REPONAME = 'multi-int-store'
+        COVERAGE_EXCLUSIONS = '**/test/**/*,**/itests/**/*,**/*Test*,**/sdk/**/*,**/*.js,**/node_modules/**/*,**/jaxb/**/*,**/wsdl/**/*,**/nces/sws/**/*,**/*.adoc,**/*.txt,**/*.xml'
```
Kjames5269: Why do we need this?
does SonarCloud do code coverage?

Jenkinsfile
```diff
@@ -61,6 +65,21 @@ pipeline {
                 }
             }
         }
+        stage('Quality Analysis - SonarCloud') {
+            when {
+                allOf {
+                    expression { env.CHANGE_ID == null }
```
Kjames5269: ❓ is this variable set anywhere or is it defaulted by checkout scm?

jhunzik: It's not set, so I believe it's checkout scm.

Jenkinsfile
jhunzik: So that we only analyze classes and not things like xml files. This was copy/pasted from replications repo, so we don't need most of them now.

PR #10, opened 2019-05-06
Removed unnecessary coverage exclusions

jhunzik:
Coverage is at 0% despite having unit tests. This PR removes the coverage exclusions that were copied over from another project. Sonarqube can only be ran on the master branch so the PR build doesn't mean anything.


jhunzik:
build now

Kjames5269:
Should we still have _some_ exclusions? like the `MultiIntStoreApplication?`

jhunzik:
@Kjames5269 We have a test for that so I think we can keep it. I'm not sure if we have anything worth excluding at this point.

PR #17, opened 2019-05-20
Add the ability to run our services in Docker containers.

jhunzik:
This PR adds docker files to each app module and a docker-compose file to start them up together. Instructions on how to do this are in the README. 

Note:
The `depends_on` property is only to check if the container came up, it doesn't ensure that the service is ready. The current workaround is to keep restarting the MIS until it connects to Cassandra.


jhunzik:
I'll need to refactor this to create a Docker stack that we can deploy to a swarm.

jhunzik:
Full local build passed.

PR #19, opened 2019-05-23
Adding Solr 

vinamartin:
Refactoring dependency hell

### DIFF
--- a/multi-int-store/build.gradle
+++ b/multi-int-store/build.gradle
@@ -4,6 +4,7 @@ plugins {
 }
 
 dependencies {
+	implementation "org.springframework.data:spring-data-solr:4.0.8.RELEASE"
 	implementation("org.springframework.data:spring-data-cassandra:2.1.5.RELEASE") {
 		exclude group: "com.google.guava", module: "guava"
 		because "CVE-2018-10237 in guava 19.0"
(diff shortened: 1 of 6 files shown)


multi-int-store/src/main/java/com/connexta/multiintstore/controllers/StorageController.java
```diff
@@ -36,21 +30,4 @@
   public StorageController(StorageManager storageManager) {
     this.manager = storageManager;
   }
-
-  @PostMapping(path = "/store/{ingestID}", consumes = "application/json")
-  public ResponseEntity store(
```
vinamartin: Oh this change shouldn't be in here, not sure why it is

multi-int-store/src/main/java/com/connexta/multiintstore/storage/persistence/config/SolrConfig.java
```diff
@@ -0,0 +1,42 @@
...
+@Configuration
+@EnableSolrRepositories(
+    basePackages = "com.connexta.multiintstore.storage.persistence.repository",
+    namedQueriesLocation = "classpath:solr-named-queries.properties")
+@ComponentScan
+public class SolrConfig {
+
+  @Bean
+  public SolrClient solrClient() {
+    return new HttpSolrClient.Builder("http://localhost:8983/solr").build();
```
vinamartin: This should be refactored out to use properties once the cassandra PR is merged

multi-int-store/build.gradle
```diff
@@ -4,6 +4,7 @@ plugins {
 }
 
 dependencies {
+	implementation "org.springframework.data:spring-data-solr:4.0.8.RELEASE"
```
Kjames5269: Out of curiosity why are we using this version of the spring framework?

vinamartin: Newest version

emmberk:
build now

PR #22, opened 2019-06-03
Update license

jhunzik:
Just updating the license manually for now. There is a ticket out to find a plugin that will allow us to verify license compatibility with our dependencies.

### DIFF
--- a/ingest/src/main/java/com/connexta/ingest/IngestApplication.java
+++ b/ingest/src/main/java/com/connexta/ingest/IngestApplication.java
@@ -1,17 +1,8 @@
 /*
- * Copyright (c) Connexta
+ * Copyright (c) 2019 Connexta, LLC
  *
- * This is free software: you can redistribute it and/or modify
- * it under the terms of the GNU Lesser General Public License as
- * published by the Free Software Foundation, either version 3
- * of the License, or any later version.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
- * GNU Lesser General Public License for more details. A copy of the
- * GNU Lesser General Public License is distributed along with this
- * program and can be found at http://www.gnu.org/licenses/lgpl.html.
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
  */
 package com.connexta.ingest;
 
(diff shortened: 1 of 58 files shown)


ingest/src/main/java/com/connexta/ingest/IngestApplication.java
```diff
@@ -1,18 +1,9 @@
...
- * of the License, or any later version.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
- * GNU Lesser General Public License for more details. A copy of the
- * GNU Lesser General Public License is distributed along with this
- * program and can be found at http://www.gnu.org/licenses/lgpl.html.
- */
+Copyright (c) 2019 Connexta, LLC
```
emmberk: ❓ Does the license need to be on non-java files too?

emmberk: Should we format the header like this instead:
```java
/*
 * Copyright (c) 2019 Connexta, LLC
 * Released under the GNU Lesser General Public License version 3; see
 */
```
That's how Apache does it.

jhunzik: This was the format that I was given.

emmberk: I'm challenging that. I will ask Eric.

jhunzik: We haven't applied the license to non-java files so far. The guidance I'm going off of says all source-controlled files, so I would think non-java files apply.

emmberk: Let's block this PR until we have answered this question because this PR isn't blocking any other tasks.

emmberk: I'm fine addressing the non-java files in another PR if you can open an issue against this repo to do that later.

emmberk: @lessarderic said he likes 
```java
/*
 * Copyright (c) 2019 Connexta, LLC
 *
 * Released under the GNU Lesser General Public License version 3; see
 */
``` better. This PR should be updated to do that instead.

jhunzik: Great! Thanks for following up on this. I'll update it.

emmberk:
build now

emmberk:
The CI build failed because of failures to download OWASP CVEs. We should set up a proxy. I'm overriding the CI build on this PR because spotless was successful.

ingest/src/main/java/com/connexta/ingest/IngestApplication.java
emmberk: https://github.com/connexta/multi-int-store/issues/26

PR #25, opened 2019-06-04
Added solr docker image

emmberk:
I still need to figure out how to test this correctly and make sure that the adaptor can talk to it. Not sure if we should hold up this PR to figure that out.


vinamartin:
build now

PR #28, opened 2019-06-05
Docker Improvements

oconnormi:
# Overview
* Removes application plugin from build.gradle files
* Configures the bootJar task to include the launch scripts
* Adds docker task
* Refactors/cleans up Dockerfiles
* Fixes up docker-compose file

# Details

## Docker Compose File
The main problem with the docker-compose file was that the `target` ports for each of the multi-int-store components were incorrect. The Dockerfiles had options built into them for `-Dserver.port=<integer>` that made it seem as though a custom port number was being used. there was nothing in place in the actual application to make use of this value however. Each of the spring-boot apps was actually binding 8080 within the docker containers. This meant that while docker was exposing the service on a custom port via the `published: 9041` for example, the target port was also 9041 _inside_ the container which was not a port that was bound by the app.

All the spring boot apps have been updated in the compose file to use `target: 8080`

## Gradle Build Changes
The three spring boot apps were configured in gradle to produce multiple outputs which weren't all necessary. This has been updated to just produce a single executable jar in all cases. This jar is a _true_ executable. This means that we don't need to do `java -jar <path to jar>`, instead we can name the file `cdr-ingest` (no .jar extension) for example and simply run ./cdr-ingest.

Additionally each docker image is now built automatically when running `./gradlew build` *This means that a docker daemon must be available to run the build*

# Testing
## Compose
* Build the project via `./gradlew build`
* Simply run `docker-compose up -d` to start all services in the background
* To check all services are up run `docker-compose ps`
* To tail the logs for all services run `docker-compose logs -f`
* To tail the logs for a subset of services run `docker-compose logs -f <serviceName> <serviceName>...`
* To clean up all services run `docker-compose down`

### Notes

The multi-int-store doesn't gracefully handle cases where cassandra isn't "ready" yet. To handle this I've added the `restart: on-failure` option to that service.

# Tasks
- [x] Add docker-compose details to README
- [ ] Add docker stack specific items

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,31 @@
 ```
 Because the Docker images are built automatically, a docker daemon must be available.
 
-## Running via docker-compose
+
+### Tests
+Tests are run automatically with `./gradlew build`. Even if the tests fail, the artifacts are built and can be run.
+
+To run tests with quiet logs, add `-Pquiet`. This should not be used with parallel builds.
+
+To run a single test suite, execute:
+```
+./gradlew module:test --test fullClassName
+```
+
+#### Integration Tests
+The integration tests require that Docker is running.
+
+To skip integration tests, add `-PskipITests`.
+
+## Running
+1. Boot up an [Apache Cassandra](https://cassandra.apache.org/) instance using the default hostname and ports: `localhost:9042`.
+
+2. To start the services, execute:
+	```
+	./gradlew run
+	```
+	
+### Running via docker-compose
 There is a docker-compose environment included in this repository.
 This will spin up each of the services that make up the multi-int-store as well as any 3rd party services needed by the multi-int-store.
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -11,6 +11,23 @@ using the default hostname and ports: `localhost:9042`
 ```
 ./gradlew run
 ```
+
+## Running via docker-compose
+There is a docker-compose environment included in this repository.
```
vinamartin: For some reason spotless is breaking on this line ->
```suggestion
There is a docker-compose environment included in this repository.
```

oconnormi: ah sorry hadn't run the build since I made the readme change. I assumed a README change wouldn't affect the build

emmberk:
build now

ahoffer:
build now

vinamartin:
@ahoffer tested this locally and deployed it. ✅ Merging

PR #32, opened 2019-06-06
Added S3 storage capability

jhunzik:


ingest/src/main/resources/application.yaml
```diff
@@ -0,0 +1,6 @@
+s3Properties:
```
jhunzik: This file isn't used and will be removed

ingest/src/main/java/com/connexta/ingest/adaptors/S3Adaptor.java
```diff
@@ -0,0 +1,68 @@
...
+import software.amazon.awssdk.services.s3.model.S3Exception;
+
+@Service
+public class S3Adaptor implements Adaptor {
+
+  private S3Client s3Client;
+
+  // We'll probably change these when we move to a cloud friendly configuration
+  private static final String S3_BUCKET = "staged-products";
+  private static final String S3_ENDPOINT = "http://127.0.0.1:9000";
```
vinamartin: If you keep the properties file, you should be able to do something like this in Spring:
```
@Value("${s3.endpoint}")
private string S3endpoint;
```

To populate it from the config file.

jhunzik: I had that originally and I can't remember why I changed it. I think we may be able to override those properties with env variables or similar when running in docker.

oconnormi: yes if you have the `@Value("${s3.endpoint}")` in your Configuration you can override it via `application.yml`, `ENV` or `--s3.endpoint=<stuff>`

oconnormi:
I think you will need to include a configuration property to allow pointing at a file containing an aws access key that can be read at runtime so that this will work with docker swarm secrets

ingest/build.gradle
```diff
@@ -13,5 +13,11 @@ docker {
 	buildArgs([JAR_FILE: "${tasks.bootJar.outputs.files.singleFile.name}"])
 }
 
+dependencies {
+	implementation platform("software.amazon.awssdk:bom:2.5.29")
+	implementation "software.amazon.awssdk:s3"
+
```
Kjames5269: What's the difference between `platform("x")` and `x`?

ingest/src/main/java/com/connexta/ingest/adaptors/S3Adaptor.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @Value("${aws.s3.endpointUrl}")
+  private String S3_ENDPOINT;
+
+  @PostConstruct
+  private void initializeS3() {
+    s3Client =
+        S3Client.builder()
+            .endpointOverride(URI.create(S3_ENDPOINT))
+            .region(Region.US_GOV_EAST_1)
```
Kjames5269: ❓ should this be customizable?

ingest/src/main/java/com/connexta/ingest/service/api/IngestRequest.java
```diff
@@ -6,9 +6,24 @@
...
  **
- * Java object representation of the json request
+ * Java object representation of the ingest request
  *
  */
-public interface IngestRequest {}
+@Getter
+@Setter
+@AllArgsConstructor
+public class IngestRequest {
```
Kjames5269: This class can be found in the Ion-Ingest-API in the `InlineObject` class. Should we be using that class instead after the Owasp errors get fixed?

ingest/src/main/java/com/connexta/ingest/adaptors/S3Adaptor.java
vinamartin: ❗️ I would think so

ingest/src/main/java/com/connexta/ingest/adaptors/S3Adaptor.java
```diff
@@ -0,0 +1,79 @@
...
+import software.amazon.awssdk.services.s3.model.S3Exception;
+
+@Service
+public class S3Adaptor implements Adaptor {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(S3Adaptor.class);
+  private S3Client s3Client;
+
+  @Autowired private String s3BucketQuarantine;
+
```
vinamartin: I think typically using `@Autowired` on the constructor and using setters is safer than inlining it here.

ingest/src/main/java/com/connexta/ingest/service/api/IngestRequest.java
```diff
@@ -6,9 +6,24 @@
...
 /*
  **
- * Java object representation of the json request
+ * Java object representation of the ingest request
  *
  */
-public interface IngestRequest {}
+@Getter
+@Setter
+@AllArgsConstructor
```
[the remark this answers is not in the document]
jhunzik: We don't need the setter, good catch.

ingest/src/main/java/com/connexta/ingest/service/api/IngestRequest.java
jhunzik: I planned on replacing it once we merged your other PR.

ingest/build.gradle
jhunzik: Maven's BOM(Bill of Materials) is a kind of platform that is used to keep the sdk versions up to date.

ingest/src/main/java/com/connexta/ingest/adaptors/S3Adaptor.java
jhunzik: I moved the S3 configuration over to the Docker side of things.

docker-compose.yml
```diff
@@ -36,6 +36,17 @@ services:
...
         condition: any
+    secrets:
+      - source: aws_s3_secret
+        target: /secrets/aws_s3_secret.sec
+      - source: aws_s3_access
+        target: /secrets/aws_s3_access.sec
+    command:
+      - "--aws.s3.secret.file=/secrets/aws_s3_secret.sec"
+      - "--aws.s3.access.file=/secrets/aws_s3_access.sec"
+      - "--aws.s3.endpointUrl=http://s3.us-gov-west-1.amazonaws.com/"
```
emmberk: ❗️remove this

docker-compose.yml
```diff
@@ -36,6 +36,17 @@ services:
...
+      - "--aws.s3.bucket.quarantine=ingest-quarantine"
+      - "--aws.s3.region=us-gov-west-1"
```
emmberk: ❗️remove this

ingest/build.gradle
```diff
@@ -27,5 +27,11 @@ docker {
 	buildArgs([JAR_FILE: "${tasks.bootJar.outputs.files.singleFile.name}"])
 }
 
+dependencies {
+	implementation platform("software.amazon.awssdk:bom:2.5.29")
+	implementation "software.amazon.awssdk:s3"
+	compile group: "commons-io", name: "commons-io", version: "2.6"
```
jhunzik: I'll remove this as it's not used anymore.

emmberk:
build now

emmberk:
The CI build failed because of a known issue downloading CVEs.

PR #37, opened 2019-06-08
Added the transform service client

ahoffer:
Created a client to talk to the transform service.

### DIFF
--- a/ingest/src/main/java/com/connexta/ingest/transform/NoOpResponseErrorHandler.java
+++ b/ingest/src/main/java/com/connexta/ingest/transform/NoOpResponseErrorHandler.java
@@ -9,7 +9,7 @@ package com.connexta.ingest.transform;
 import org.springframework.http.client.ClientHttpResponse;
 import org.springframework.web.client.DefaultResponseErrorHandler;
 
-// Do not throw an exception when an HTTP response include a non-successful status code
+// Do not throw an exception when an HTTP response includes a non-successful status code
 public class NoOpResponseErrorHandler extends DefaultResponseErrorHandler {
 
   @Override
--- a/ingest/src/main/java/com/connexta/ingest/transform/NoOpResponseErrorHandler.java
+++ b/ingest/src/main/java/com/connexta/ingest/transform/NoOpResponseErrorHandler.java
@@ -1,0 +1,17 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.ingest.transform;
+
+import org.springframework.http.client.ClientHttpResponse;
+import org.springframework.web.client.DefaultResponseErrorHandler;
+
+// Do not throw an exception when an HTTP response include a non-successful status code
+public class NoOpResponseErrorHandler extends DefaultResponseErrorHandler {
+
+  @Override
+  public void handleError(ClientHttpResponse response) {}
+}
--- a/ingest/src/main/java/com/connexta/ingest/transform/TransformClient.java
+++ b/ingest/src/main/java/com/connexta/ingest/transform/TransformClient.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.ingest.transform;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Service;
+import org.springframework.web.client.ResponseErrorHandler;
+import org.springframework.web.client.RestTemplate;
+
+@Service
+public class TransformClient {
+
+  @Autowired private RestTemplate restTemplate;
+
+  private URI transformEndpoint;
+
+  public TransformResponse requestTransform(TransformRequest transformRequest) {
+    ResponseErrorHandler originalRequestHandler = restTemplate.getErrorHandler();
+    try {
+      restTemplate.setErrorHandler(new NoOpResponseErrorHandler());
+      return postForTransform(transformRequest);
+    } finally {
+      restTemplate.setErrorHandler(originalRequestHandler);
+    }
+  }
+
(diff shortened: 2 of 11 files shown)


ingest/src/main/java/com/connexta/ingest/transform/TransformClient.java
```diff
@@ -0,0 +1,63 @@
...
+public class TransformClient {
+
+  @Autowired private RestTemplate restTemplate;
+
+  private URI transformEndpoint;
+
+  public TransformResponse requestTransform(TransformRequest transformRequest) {
+    ResponseErrorHandler originalRequestHandler = restTemplate.getErrorHandler();
+    TransformResponse transformResponse = null;
+    try {
```
vinamartin: Can you clarify what this syntax means? What happens if an exception gets thrown in the try? Does it get bubbled up somewhere?

ingest/src/test/java/com/connexta/ingest/transform/TransformClientTest.java
```diff
@@ -0,0 +1,116 @@
...
+  private MockRestServiceServer mockServer;
+  private String transformEndpoint;
+  private TransformResponse expectedSuccessResponse;
+  private TransformRequest transformRequest;
+  private TransformResponse expectedFailedResponse;
+
+  @Before
+  public void setUp() throws URISyntaxException {
+    mockServer = MockRestServiceServer.createServer(restTemplate);
+    transformEndpoint = "http://localhost:8000/transform";
```
vinamartin: Is this url just a test one that you don't actually use while the test is running?
If you are actually using the url in the test (which I'm not 100% clear on), there's a way to get the actual running container host and port using spring.

ingest/src/main/java/com/connexta/ingest/transform/NoOpResponseErrorHandler.java
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.ingest.transform;
+
+import org.springframework.http.client.ClientHttpResponse;
+import org.springframework.web.client.DefaultResponseErrorHandler;
+
+// Do not throw an exception when an HTTP response include a non-successful status code
```
Kjames5269: ✏️`includes`

ingest/src/test/java/com/connexta/ingest/transform/TransformClientTest.java
```diff
@@ -0,0 +1,117 @@
...
+    expectedFailedResponse.setDetails(List.of("A", "B"));
+  }
+
+  @Test
+  public void testSuccessfulRequest() throws JsonProcessingException {
+
+    mockServer
+        .expect(requestTo(transformEndpoint))
+        .andExpect(method(HttpMethod.POST))
+        .andExpect(jsonPath("$.id").value("1"))
```
Kjames5269: ✏️ `.andExpect(jsonPath("$.id").value(transformRequest.getId())`
. . .

ingest/src/main/java/com/connexta/ingest/transform/TransformRequest.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+package com.connexta.ingest.transform;
+
+import java.io.Serializable;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+public class TransformRequest implements Serializable {
```
emmberk: ✏️Use a constructor and final fields instead of setters so that we can enforce that the values are never null and shouldn't be changed

ingest/src/main/java/com/connexta/ingest/transform/TransformResponse.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import java.util.List;
+import lombok.Getter;
+import lombok.Setter;
+import org.springframework.http.HttpStatus;
+
+@Getter
+@Setter
+public class TransformResponse {
```
emmberk: ✏️Use a constructor and final fields instead of setters so that we can enforce that the values are never null and shouldn't be changed

emmberk:
I tested ./gradlew clean build locally

emmberk:
build now

ingest/src/test/java/com/connexta/ingest/transform/TransformClientTest.java
ahoffer: It's a test that should use actual HTTP requests and responses, but the mock server will actually service the requests. The mock server only exists for the lifetime of the test. It is not meant to invoke services provided by another container.

ingest/src/main/java/com/connexta/ingest/transform/TransformRequest.java
```diff
@@ -0,0 +1,69 @@
...
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.ingest.transform;
+
+import java.io.Serializable;
+
+public class TransformRequest implements Serializable {
```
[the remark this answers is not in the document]
emmberk: Fixed

ingest/src/main/java/com/connexta/ingest/transform/NoOpResponseErrorHandler.java
emmberk: Fixed

ingest/src/main/java/com/connexta/ingest/transform/TransformClient.java
ahoffer: It turns that the default exception handler throws exception for reponse status codes outside of the 200 range. That isn't what we want to do because it would be hard to recover the JSON details/messages in the responses for malformed requests and server errors. I install a no-op error handler so I can extract an JSON the server sends in the response.

I let Spring the RestTemplate object, which is shared with other classes in the environment  so I don't want to replace the default error handler for everyone. I temporarily replace the error handler. 

There is probably a more elegant way to do this.

ingest/src/test/java/com/connexta/ingest/transform/TransformClientTest.java
ahoffer: In unit tests I usually test for the actual values. I've had too many cases where the value was not set on the object and that hit an error in the code.

emmberk:
The CI buildFailed to download CVEs. This is a known issue. I'm going to override this build check.

PR #38, opened 2019-06-10
Adds tests for metadata callback and fetches CST

vinamartin:
Once we actually retrieve the metadata, assert it's correct


multi-int-store/src/test/java/com/connexta/multiintstore/MultiIntStoreIntegrationTest.java
```diff
@@ -31,6 +41,22 @@ public void setup() throws Exception {
...
+            .content("{\n" + "    \"id\": \"1234\",\n" + "    \"status\": \"COMPLETE\",\n"
+                    + "    \"type\": \"ddms5.0\",\n" + "    \"mimeType\": \"application/xml\",\n"
+                    + "    \"bytes\": \"256\",\n"
+                    + "    \"location\": \"https://localhost:8080\",\n" + "    \"security\": {\n"
+                    + "        \"classification\": \"U\",\n"
+                    + "        \"ownerProducer\": \"ownerProducer\"\n" + "    }\n" + "}"))
+            .andExpect(status().isOk());
+    List<Metadata> metadatas = new ArrayList<>();
+    repository.findAll().forEach(metadatas::add);
+    assertThat(metadatas.size(), is(1));
```
emmberk: ✏️Nit: Seems like there is an easier way to verify the size than copying to a new List, e.g. `assertThat(Iterables.size(repository.findAll()), is(1));` or a hamcrest matcher.

Kjames5269: Once we actually get real metadata to store we'll find it by the real ID, but for now I think this _should_ be fine.

emmberk:
build now

PR #40, opened 2019-06-11
Adding itests for the ingest api

Kjames5269:


Kjames5269:
build now

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
```diff
@@ -6,13 +6,58 @@
...
+
+    mvc.perform(multipart("/ingest").file("file", "some-content".getBytes())
+                    .param("fileSize", "10")
+                    .param("filename", "file")
+                    .param("title", "qualityTitle")
+                    .param("mimeType", "plain/text")
+                    .header("Accept-Version", "1.2.1")
+                    .accept(MediaType.APPLICATION_JSON)
+                    .contentType(MediaType.MULTIPART_FORM_DATA))
+        .andExpect(status().isNotImplemented());
```
jhunzik: ❗️ Valid requests return a 202

ingest/build.gradle
```diff
@@ -4,9 +4,13 @@ plugins {
 }
 
 dependencies {
-	testCompile group: "org.springframework.boot", name: "spring-boot-starter-test", version: "2.1.5.RELEASE"
```
jhunzik: ✏️ Could we group the types of dependencies?

ingest/src/main/java/com/connexta/ingest/adaptors/S3Adaptor.java
```diff
@@ -52,9 +51,10 @@ public S3Adaptor(S3StorageConfiguration configuration) {
     s3Endpoint = configuration.getS3Endpoint();
     s3Region = configuration.getS3Region();
     s3BucketQuarantine = configuration.getS3BucketQuarantine();
+
+    initializeS3();
```
jhunzik: It's weird that we have to do this because that's what `@PostConstruct` is supposed to be doing. 😞

ingest/build.gradle
Kjames5269: that seems like a reasonable request

ingest/src/main/java/com/connexta/ingest/adaptors/S3Adaptor.java
Kjames5269: I agree, I think we can actually move back to `@PostConstruct`

ahoffer:
Heroed commit f85.
Passed the build tasks with tests.

However, the ingest container could not be started because of an uninitialized value for an endpoint on line 53 of S3Adaptor.
`Caused by: java.lang.NullPointerException: The URI scheme of endpointOverride must not be null.`

I think this PR should be merged despite that. Master branch won't work in the cloud because many values are uninitialized and holding up this PR is could make rebasing and merging more difficult in the short term.

PR #42, opened 2019-06-11
Query endpoint and search service

ahoffer:
This PR:
1. Adds a **QueryController** to create a */search* endpoint.
1. Adds a **SearchService** to interact with the Solr **SearchRepository**.
1. Adds integration test to test the /search endpoint.
1. Renames some classes and packages to tighten up the names and match Spring idioms for naming services. 
    1. The most important changes are to classes. The *CommonSearchTerms* class is now the **SearchDocument**. The *CommonSearchTermsRepository* is now the **SearchRepository**. 
    1. Because of the name changes, some merging will have to take place with other work in progress when it is time to merge. I'll volunteer to do reconcile any code, just let me know when it's time.


multi-int-store/build.gradle
```diff
@@ -1,10 +1,11 @@
 plugins {
 	id "org.springframework.boot" version "2.1.5.RELEASE"
 	id "com.palantir.docker" version "0.22.1"
+	id "io.spring.dependency-management" version "1.0.7.RELEASE"
```
vinamartin: This particular plugin has caused a lot of headaches for us in the past (by pulling in non-compatible versions through transitive dependencies). If we think it's fine to bring it back in we should be making this decision for the entire code base instead of individual dependencies.

multi-int-store/src/main/java/com/connexta/multiintstore/controllers/QueryController.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping(value = "/search")
+public class QueryController {
+
+  @Autowired private SearchService searchService;
```
vinamartin: Move this into an autowired constructor 
```suggestion
private SearchService searchService;

@Autowired
public QueryController (SearchService searchService) {
    this.searchService = searchService;
}
```

multi-int-store/src/main/java/com/connexta/multiintstore/repositories/SearchRepository.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+package com.connexta.multiintstore.repositories;
+
+import com.connexta.multiintstore.models.SearchDocument;
+import java.util.List;
+import org.springframework.data.solr.repository.SolrCrudRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface SearchRepository extends SolrCrudRepository<SearchDocument, String> {
```
vinamartin: For consistency, wouldn't this be better named "SearchDocumentRepository"?

vinamartin:
✏️ Relatively minor nitpick: I'm still unsure on the naming "SearchDocument" since it's more of a metadata that represents something you'd have to retrieve somewhere else. That said, I can't think of anything better so I don't think it needs to be changed if no one else has an issue with it.

multi-int-store/build.gradle
ahoffer: I'll pull it out and see what happens.

multi-int-store/src/main/java/com/connexta/multiintstore/repositories/SearchRepository.java
ahoffer: Of the examples I've been looking at on the web, one put the word Document before Repository:
  https://www.petrikainulainen.net/programming/solr/spring-data-solr-tutorial-crud-almost/

and four did not:
  https://www.baeldung.com/spring-data-solr
  https://github.com/christophstrobl/spring-data-solr-showcase/tree/master/src/main/java/org/springframework/data/solr/showcase/product
  https://dzone.com/articles/crud-application-using-spring-data-solr-and-spring
  https://github.com/spring-projects/spring-data-solr-examples/tree/master/spring-solr-repository-example/src/main/java/org/springframework/data/solr/example/repository

I like the shorter "SearchRepository", but I would be OK with "SearchDocumentRepository"

ahoffer:
>  Relatively minor nitpick: I'm still unsure on the naming "SearchDocument" since it's more of a metadata that represents something you'd have to retrieve somewhere else. That said, I can't think of anything better so I don't think it needs to be changed if no one else has an issue with it.

I've struggled with the name too. I never liked metadata because it is such an over loaded term. "Search" is overloaded as well, but I liked it because it spoke to the purpose of the (meta) data. Maybe **ProductMetadata** is a good choice? I like that that. @vinamartin Does **ProductMetadataDocument** sound better than SearchDocument? 

But then how does that distinguish this metadata from the metadata to be stored in Cassandra? Maybe we drop the word word "Document" entirely? Go with **SearchableMetadata** or **IndexedMetadata**? I could go with either of those too.

vinamartin:
@ahoffer ProductMetadata is not too bad- I could also see DocumentMetadata working as well.
What do others think?

ahoffer:
> @ahoffer ProductMetadata is not too bad- I could also see DocumentMetadata working as well.
> What do others think?

I like ProductMetadata because it is more specific. "Product" is a term from the domain and we use it in other places.

ahoffer:
I went with `IndexedMetadataRepository`. @vinamartin what do you think?

ahoffer:
Problem running itests. StorageManager is looking for, but cannot find a ProductDao bean.

```rror creating bean with name 'storageManager' defined in file [/projects/multi-int-store/multi-int-store/build/classes/java/main/com/connexta/multiintstore/common/StorageManager.class]: Unsatisfied dependency expressed through constructor parameter 1; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'com.connexta.multiintstore.services.api.Dao<com.connexta.multiintstore.models.IndexedProductMetadata, java.lang.String>' available: expected at least 1 bean which qualifies as autowire candidate.```

**Comment by ahoffer** [comment]
build now

File: multi-int-store/src/main/java/com/connexta/multiintstore/repositories/IndexedMetadataRepository.java [hunk 4]
```diff
@@ -0,0 +1,19 @@
...
+import com.connexta.multiintstore.models.IndexedProductMetadata;
+import java.util.List;
+import org.springframework.data.solr.repository.SolrCrudRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface IndexedMetadataRepository
+    extends SolrCrudRepository<IndexedProductMetadata, String> {
+
+  List<IndexedProductMetadata> findByContents(String keyword);
```
vinamartin: ❓ Is this method built in and already handles the query for us? I'm a little confused by this syntax.

vinamartin: ❓ Is this method built in and already handles the query for us? I'm a little confused by this syntax.

multi-int-store/src/main/java/com/connexta/multiintstore/controllers/IndexingController.java
```diff
@@ -6,17 +6,13 @@
...
+@AllArgsConstructor
 public class IndexingController {
 
-  private final CommonSearchTermsRepository commonSearchTermsRepository;
-
-  @Autowired
-  public IndexingController(CommonSearchTermsRepository commonSearchTermsRepository) {
-    this.commonSearchTermsRepository = commonSearchTermsRepository;
-  }
+  private final IndexedMetadataRepository commonSearchTermsRepository;
```
emmberk: ❓Why is this no longer `Autowired`?

multi-int-store/src/main/java/com/connexta/multiintstore/services/impl/SearchServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+import com.connexta.multiintstore.services.api.SearchService;
+import java.util.List;
+import lombok.AllArgsConstructor;
+import org.springframework.stereotype.Service;
+
+@Service
+@AllArgsConstructor
+public class SearchServiceImpl implements SearchService {
+
+  private IndexedMetadataRepository indexedMetadataRepository;
```
emmberk: ✏️ `final`

multi-int-store/src/main/java/com/connexta/multiintstore/controllers/QueryController.java
```diff
@@ -0,0 +1,31 @@
...
+@RestController
+@RequestMapping(value = "/search")
+@AllArgsConstructor
+public class QueryController {
+
+  private SearchService searchService;
+
+  @RequestMapping(method = RequestMethod.GET, params = "q")
+  @ResponseBody
+  public List<IndexedProductMetadata> searchKeyword(@RequestParam(value = "q") String keyword) {
```
emmberk: ✏️ Do the results need to be ordered? Can we use a `Set` instead?

multi-int-store/src/test/java/com/connexta/multiintstore/MultiIntStoreIntegrationTest.java
```diff
@@ -32,7 +34,8 @@
 public class MultiIntStoreIntegrationTest extends MultiIntStoreIntegrationTestContainers {
 
   @Autowired private WebApplicationContext wac;
-  @Autowired private CommonSearchTermsRepository cstRepository;
+  @Autowired private IndexedMetadataRepository searchRepository;
+  @Autowired private SearchService searchService;
```
emmberk: ✏️ I know this is a test class, but we can use an `Autowired` constructor for consistency
```java
private final WebApplicationContext wac;
private final IndexedMetadataRepository searchRepository;
private final SearchService searchService;

@Autowired
public MultiIntStoreIntegrationTest(WebApplicationContext wac, IndexedMetadataRepository searchRepository, SearchService searchService) {
  this.wac = wac;
  this.searchRepository = searchRepository;
  this.searchService = searchService;
}
```

multi-int-store/src/test/java/com/connexta/multiintstore/MultiIntStoreIntegrationTest.java
```diff
@@ -93,4 +96,13 @@ public void testRetrieve() throws Exception {
         .andDo(print())
         .andExpect(status().isOk());
   }
+
+  @Test
+  public void testSearchEndpointKeyword() throws Exception {
```
emmberk: Should this method also verify `searchService.find`?
```java
final IndexedProductMetadata indexedProductMetadata =
    new IndexedProductMetadata(
        "1",
        "All the color had been leached from Winterfell until only grey and white remained");
searchService.store(indexedProductMetadata);

assertThat(searchService.find("Winterfell"), contains(indexedProductMetadata));
```

multi-int-store/src/test/java/com/connexta/multiintstore/MultiIntStoreIntegrationTest.java
```diff
@@ -93,4 +96,13 @@ public void testRetrieve() throws Exception {
...
+
+    searchService.store(
+        new IndexedProductMetadata(
+            "1",
+            "All the color had been leached from Winterfell until only grey and white remained"));
```
emmberk: 😆

multi-int-store/src/main/java/com/connexta/multiintstore/common/StorageManager.java
```diff
@@ -24,13 +24,13 @@
   private static final String CST = "cst";
```
jhunzik: ❓ Should this variable name change now that we call it `IndexedProductMetadata`?

multi-int-store/src/main/java/com/connexta/multiintstore/controllers/IndexingController.java
ahoffer: The Lombok `@AllArgsConstructor` autowires the constructor.

multi-int-store/src/main/java/com/connexta/multiintstore/controllers/QueryController.java
ahoffer: Solr will order the results based on a relevancy score.

multi-int-store/src/main/java/com/connexta/multiintstore/repositories/IndexedMetadataRepository.java
ahoffer: The magic of Spring data does the... magic. It will look for a Solr field named "contents". 
https://docs.spring.io/spring-data/data-solr/docs/1.0.0.RC1/reference/htmlsingle/#solr.query-methods.criterions

multi-int-store/src/test/java/com/connexta/multiintstore/MultiIntStoreIntegrationTest.java
ahoffer: Yes, I think it should. That is what the method name purports to test.

multi-int-store/src/test/java/com/connexta/multiintstore/MultiIntStoreIntegrationTest.java
```diff
@@ -13,26 +13,31 @@
...
 import org.springframework.web.context.WebApplicationContext;
 
+@AllArgsConstructor
 public class MultiIntStoreIntegrationTest extends MultiIntStoreIntegrationTestContainers {
 
-  @Autowired private WebApplicationContext wac;
-  @Autowired private CommonSearchTermsRepository cstRepository;
+  private WebApplicationContext wac;
+  private IndexedMetadataRepository searchRepository;
+  private SearchService searchService;
```
emmberk: ✏️ `final`

multi-int-store/src/test/java/com/connexta/multiintstore/MultiIntStoreIntegrationTest.java
```diff
@@ -93,4 +98,19 @@ public void testRetrieve() throws Exception {
...
+    searchService.store(
+        new IndexedProductMetadata(
+            "1",
+            "All the color had been leached from Winterfell until only grey and white remained"));
+
+    List<IndexedProductMetadata> results = searchService.find("Winterfell");
+
+    searchService.delete("1");
+
+    results = searchService.find("Winterfell");
```
emmberk: Shouldn't the results be verified?

multi-int-store/src/main/java/com/connexta/multiintstore/common/StorageManager.java
```diff
@@ -70,7 +70,7 @@ private void handleCallback(MetadataCallback callback, String ingestId) {
       String contents =
           retriever.getMetadata(
               callback.getLocation().toString(), callback.getMimeType(), String.class);
-      final CommonSearchTerms cst = new CommonSearchTerms(ingestId, contents);
+      final IndexedProductMetadata cst = new IndexedProductMetadata(ingestId, contents);
```
emmberk: ✏️ update variable name as well

multi-int-store/src/test/java/com/connexta/multiintstore/MultiIntStoreIntegrationTest.java
emmberk: ✏️ Verify the retrieve endpoint, not the SearchService

multi-int-store/src/main/java/com/connexta/multiintstore/services/api/SearchService.java
```diff
@@ -0,0 +1,19 @@
...
+import com.connexta.multiintstore.models.IndexedProductMetadata;
+import java.util.List;
+
+public interface SearchService {
+
+  void store(IndexedProductMetadata doc);
+
+  List<IndexedProductMetadata> find(String keyword);
+
+  void delete(String id);
```
emmberk: ✏️ We can probably delete this because we don't/won't use it

multi-int-store/src/main/java/com/connexta/multiintstore/common/StorageManager.java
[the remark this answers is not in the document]
emmberk: No. `cst` is still a special type of metadata. We use this variable to check what type of metadata is in the callback.

multi-int-store/src/test/java/com/connexta/multiintstore/MultiIntStoreIntegrationTest.java
```diff
@@ -100,17 +100,85 @@ public void testRetrieve() throws Exception {
...
-    searchService.delete("1");
+    whenHttp(server)
+        .match(Condition.endsWithUri("/location/cst001"))
+        .then(
+            Action.contentType(MediaType.TEXT_PLAIN.toString()),
+            Action.stringContent(contents),
+            Action.status(HttpStatus.OK_200));
 
-    results = searchService.find("Winterfell");
+    mockMvc.perform(
```
ahoffer: At some point we will need to think about cleaning out our persistent stores like Solr and Cassandra between itests.

emmberk: See line 49

ahoffer: I see that method on the interface. 
https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/repository/CrudRepository.html

emmberk:
build now

emmberk:
The CI build failed because of a known issue downloading CVEs. I'm overriding the CI build because OWASP was tested locally.

PR #46, opened 2019-06-13
Adding a deploy script to run for deploying docker files to the cloud

Kjames5269:
Deploy script


deploy.bash
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env bash
+
+# No failures allowed
```
vinamartin: failure is always an option

Kjames5269: :joy:

deploy.bash
```diff
@@ -0,0 +1,87 @@
...
+for i in $(echo ${listOfImages}); do
+    run "docker tag ${i} ${DEPLOY_IP}:${PORT}/${i}"
+    run "docker push ${DEPLOY_IP}:${PORT}/${i}"
+done
+
+header "Pulling and tagging docker images on ${DEPLOY_IP}:${PORT}"
+
+for i in $(echo ${listOfImages}); do
+    run "${DOCKER_W} pull ${DEPLOY_IP}:${PORT}/${i}"
+    run "${DOCKER_W} tag ${DEPLOY_IP}:${PORT}/${i} ${i}"
```
emmberk: We don't need to tag after pulling if we use this docker deploy command:
```bash
${DOCKER_W} stack deploy -c <(REGISTRY=${DEPLOY_IP}:${PORT} docker-compose -f docker-compose.yml config) cdr-stack
```

deploy.bash
```diff
@@ -0,0 +1,87 @@
...
+DOCKER_W=""
+
+GITHUB_PATCH=""
+
+# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+
+# Tests that variables are correct for the docker compose and docker wrapper
+
+if [ -z ${DEPLOY_IP} ] && [ -z ${PORT} ] && [ -z ${DOCKER_W} ]; then
+    echo "Set the global variables at the top of $0"
```
emmberk: ❓ Is there a way we can set these variables when we run the script so that I don't have to modify this file? Something like `--stackName ... --deployIp ...`.

ahoffer: I think you can do that with  the Bash version of getopts
https://www.computerhope.com/unix/bash/getopts.htm

deploy.bash
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env bash
+
+# No failures allowed
+set -e
+
+# The name of the preface for the docker images
+DOCKER_NAME="cnxta/cdr-"
+
+# The IP For the place you're storing this
+DEPLOY_IP=""
```
ahoffer: The deploy address IP can also be hostname. 
The IP/hostname and port belong together. 
These values are used for the tag and push to a Docker registry, which is not necessarily where it will be deployed.
Make one variable, something like `DOCKER_REGISTRY`. 
You don't have to define it in this file. Let users sets it in their .profile or .bashrc file like so:
`export DOCKER_REGISTRY=myreg.com:4545`

deploy.bash
Kjames5269: I decided to make it so you can have either variables in your .bash_profile or edit the file. I don't think there's a need to accept parameters as not much is going to change between deployments

Kjames5269:
I hero'd it and it worked successfully!

Kjames5269:
build now

Kjames5269:
build now

PR #47, opened 2019-06-14
Externalize the configurations

jhunzik:


multi-int-store/src/test/java/com/connexta/multiintstore/MultiIntStoreIntegrationTest.java
```diff
@@ -65,9 +65,6 @@ public void handleCSTCallback() throws Exception {
             Action.stringContent(cstContents),
             Action.status(HttpStatus.OK_200));
 
-    // TODO
-    Thread.sleep(5000);
-
```
emmberk: 🎉

README.md
```diff
@@ -171,6 +171,14 @@ docker stack deploy -c docker-compose.yml cdr
 	docker stack rm cdr-stack
 	```
 
+## Configuration
+Services can be configured with an external configuration file that will be applied to the docker container during
+deployment. The configuration YAML files can be found under: `<PROJECT_ROOT>/configs/` and are not verison-controlled.
+The properties in these files will be merged with any properties that you have configured in the service. The properties
+in the external config file take precedence over config files that are built with the service.
+
+Currently only the Ingest Service is configured externally.
```
emmberk: ✏️ Could you add an example of the contents of ingest_config.yml or something like that so I know how to format it?

jhunzik:
> How do we use the config file? We do we also need changes to the ApplicationConfiguration class?

@ahoffer I'll add an example to the README.

README.md
ahoffer: @jhunzik Could we add those file and put a note in them that they are for configuration information that should not be checked into a repo? It feels like a dangling pointer to reference a file and not provide somekind of placeholder.

emmberk:
I rebased.

README.md
```diff
@@ -180,3 +186,18 @@ S3 bucket name, and the credentials the Ingest Service will use to connect to S3
...
 controlled.
+
+Example configuration file:
+```yaml
+aws:
+s3:
+	endpointUrl: https://s3.us-west-1.amazonaws.com
+	region: us-west-1
+	bucket:
+	quarantine: ingest-qurantine
```
emmberk: `quarantine`

README.md
```diff
@@ -180,3 +186,18 @@ S3 bucket name, and the credentials the Ingest Service will use to connect to S3
...
+transformation:
+transformEndpointUrl: http://localhost:1231/transform/
+ingest:
+callbackEndpointUrl: http://localhost.com:1232/store/
+retrieveEndpointUrl: http://localhost.com:1233/retrieve/
```
emmberk: ✏️ The indenting here could be improved.

emmberk: ✏️ Remove `.com` for consistency

emmberk: I wasn't able to easily fix this. Spotless either removes the spaces or replaces them with tabs, even though yaml doesn't allow tabs. One suggestion is to link to an example yaml file.

docker-compose.yml
```diff
@@ -97,4 +102,9 @@ secrets:
...
     file: aws_s3_secret.sec
   aws_s3_access:
-    file: aws_s3_access.sec
\ No newline at end of file
+    file: aws_s3_access.sec
+configs:
+  ingest_config:
+    file: ./configs/ingest_config.yml
+  mis_config:
+    file: ./configs/mis_config.yml
```
emmberk: ❓Seems weird that we need to repeat the path to each config file three times

docker-compose.yml
```diff
@@ -63,6 +63,11 @@ services:
     deploy:
       restart_policy:
         condition: any
+    configs:
+      - source: mis_config
```
emmberk: ✏️ Should `mis_config` also contain the solr port?

emmberk: ✏️ Should
```yaml
solr:
  host: TODO
  port: 9983
```
be removed from /multi-int-store/src/main/resources/application.yml?

emmberk:
build now

docker-compose.yml
jhunzik: The paths under the `configs` section are local and the the `target` paths refer to the containers.

vinamartin:
Tested ✅

emmberk:
I rebased and added a commit to address the indenting in the example configs in the readme.

`./gradlew clean build` and `./gradlew dependencyCheckAnalyze --info` succeed for me locally.

build now

README.md
emmberk: See https://github.com/connexta/multi-int-store/pull/47/commits/4f8e16025793c2c5c78d12b6272d2821aa0c0734.

PR #50, opened 2019-06-17
Inject the callback and transform URLs

jhunzik:


ingest/src/main/java/com/connexta/ingest/service/impl/IngestServiceImpl.java
```diff
@@ -28,13 +31,21 @@
...
     this.s3Adaptor = s3Adaptor;
     this.transformClient = transformClient;
   }
 
+  @Autowired
+  @Value("${endpointUrl.ingest.callback}")
+  public void setCallbackUrl(String callbackEndpoint) throws URISyntaxException {
+    LOGGER.info("Callback endpoint URL is: {}", callbackEndpoint);
+    this.callbackUrl = callbackEndpoint;
+  }
```
emmberk: ✏️ Would it be safer to make `callbackUrl` and `transformEndpoint` final so that that are never null?

jhunzik:
build now, because why not?

Kjames5269:
Hero'd ✅

build.gradle
```diff
@@ -1,192 +1,192 @@
...
-			removeUnusedImports()
-			googleJavaFormat()
-			trimTrailingWhitespace()
-		}
-	}
+    static Closure getMisc() {
+        return {
+            target "**/*.gradle", "**/*.md", "**/.gitignore"
+            trimTrailingWhitespace()
+            indentWithSpaces(4)
```
ahoffer: :+1:

jhunzik:
Manually hero'd and tested so I'm going to override CI

PR #51, opened 2019-06-18
Gradle++

Kjames5269:
much needed Gradle quality of life improvements.

### DIFF
--- a/buildSrc/src/main/groovy/Dependencies.groovy
+++ b/buildSrc/src/main/groovy/Dependencies.groovy
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+/* Default Package */
+class Dependencies {
+
+  //  Dependencies
+  static def awsBom = { it=Versions.awsBom -> "software.amazon.awssdk:bom:${it}" }
+  static def awsS3 = { "software.amazon.awssdk:s3" }
+  static def commonsIO = { it=Versions.commonsIO -> "commons-io:commons-io:${it}" }
+  static def guava = { it=Versions.guava -> "com.google.guava:guava:${it}" }
+  static def ingestAPI = { it=Versions.ingestAPI -> "com.connexta.ingest:ingest-api-rest-spring-stubs:${it}" }
+  static def jacksonDatabind = { it=Versions.jacksonCore -> "com.fasterxml.jackson.core:jackson-databind:${it}" }
+  static def json = { it=Versions.json -> "org.json:json:${it}" }
+  static def lombok = { it=Versions.lombok -> "org.projectlombok:lombok:${it}" }
+  static def springBootWebfluxStarter = { it=Versions.springBoot -> "org.springframework.boot:spring-boot-starter-webflux:${it}" }
+  static def springBootWebStarter = { it=Versions.springBoot -> "org.springframework.boot:spring-boot-starter-web:${it}" }
+  static def springFrameworkSolr = { it=Versions.springData -> "org.springframework.data:spring-data-solr:${it}" }
+  static def tomcatCore = { it=Versions.tomcat -> "org.apache.tomcat.embed:tomcat-embed-core:${it}" }
+  static def tomcatWebsocket = { it=Versions.tomcat -> "org.apache.tomcat.embed:tomcat-embed-websocket:${it}" }
+  static def transformAPI = { it=Versions.transformAPI -> "com.connexta.transformation:transformation-api-rest-spring-stubs:${it}" }
+
+  //  Test Dependencies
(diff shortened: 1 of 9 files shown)


buildSrc/src/main/groovy/Versions.groovy
```diff
@@ -0,0 +1,35 @@
...
+// Default Package
+class Versions {
+    static String java = "1.11"
+    static String project = "0.1.0-SNAPSHOT"
+
+    //  Versions
+    static String awsBom = "2.5.29"
+    static String commonsIO = "2.6"
+    static String guava = "27.1-jre"
+    static String ingestAPI = "0.1.1"
```
jhunzik: ✏️ Maybe the API versions should be in their own group.

buildSrc/src/main/groovy/Dependencies.groovy
```diff
@@ -0,0 +1,37 @@
+/*
+Copyright (c) 2019 Connexta, LLC
+
+Released under the GNU Lesser General Public License version 3; see
+https://www.gnu.org/licenses/lgpl-3.0.html
+*/
+// Default Package
+class Dependencies {
```
emmberk: ❓Why not just stick the version right in the dependency definition for these that don't share a version with any other dependency?

emmberk: ✏️ Maybe split this into three classes: `Dependencies`, `TestDependencies`, and `PluginDependencies` so that it is obvious when we use them, e.g. `testImplementation(TestDependencies.springBootTestStarter())`.

Kjames5269: Mostly for consistency reasons. It'd be weird to be able to override a version on some dependencies but not all of them.

for example if this works
`Dependencies.foo("1.2.3")`
but this doesn't
`Dependencies.bar("1.2.3")`
I feel that'd be werid

buildSrc/src/main/groovy/Versions.groovy
Kjames5269: I feel that's not really needed since they're just another dependency, but I don't feel too strongly on that

buildSrc/src/main/groovy/Dependencies.groovy
Kjames5269: I'm not sure we need to be that obvious about dependencies we're using, I think having the `testImplementation` should be enough

emmberk: I don't understand your response. I was suggesting doing this:
```groovy
static def commonsIO = "commons-io:commons-io:2.6"
```
```gradle
implementation(Dependencies.commonsIO())
```

Kjames5269: Well there's two things here to describe, the first is the use of closures and the second is the having the version in the same spot.

the reason why I had it setup with  
```
static def commonsIO = { it=Versions.commonsIO -> "commons-io:commons-io:${it}" }
```
was to allow someone to override a version for a hopefully good reason.
`Dependencies.commonIO("1.2.3")`
would use 1.2.3 instead of the actual version.

For the second part, personal preference and consistency so all the versions are together. I think it'd look strange to have some versions be variables and some be strings

emmberk: Not a blocker for this PR. You wouldn't ever use `Versions.commonsIO` outside of `Dependencies.gradle`, right? Moving the version to `Dependencies.gradle` would enforce that.

Kjames5269: I did some fiddling with the groovy files and it doesn't look like there's a way to make something truly private in groovy. If it's private you can still use it in another file, but intellij will show a warning saying you're exceeding the access rights

emmberk: Why would you have to make it private. Would this work?
```groovy
static def commonsIO = { it="2.6" -> "commons-io:commons-io:${it}" }
```
This would maintain the ability to overwrite the version if you wanted to, right?
```gradle
implementation(Dependencies.commonIO("1.2.3"))
```

Kjames5269: I was going trying to make the variables private to keep them around. Doing that does work but I don't like how it looks since we'd be mixing strings and variables. That's my take on it
```
  static def commonsIO = { it="2.6" -> "commons-io:commons-io:${it}" }
  static def springBootWebStarter = { it=Versions.springBoot -> "org.springframework.boot:spring-boot-starter-web:${it}" }
```

Kjames5269:
build now

emmberk:
build now

emmberk:
`./gradlew clean build` and `./gradlew dependencyCheckAnalyze --info` succeeded for me locally.

emmberk:
The CI build failed because of a known issue downloading CVEs. I'm overriding the CI check because OWASP was tested locally.

emmberk:
:+1:Heroed

PR #54, opened 2019-06-18
Updated search endpoint to return retrieve URLs

emmberk:
- Injected the retrieve endpoint into the ingest service
- Updated the /search endpoint to return a list of retrieve URLs instead of the id and cst

### DIFF
--- a/ingest/src/main/java/com/connexta/ingest/service/impl/IngestServiceImpl.java
+++ b/ingest/src/main/java/com/connexta/ingest/service/impl/IngestServiceImpl.java
@@ -15,6 +15,7 @@ import com.connexta.transformation.rest.models.TransformResponse;
 import java.io.IOException;
 import java.net.URL;
 import java.util.UUID;
+import javax.validation.constraints.NotEmpty;
 import javax.validation.constraints.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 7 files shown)


emmberk:
rebased

ingest/src/main/java/com/connexta/ingest/service/impl/IngestServiceImpl.java
```diff
@@ -29,16 +30,20 @@
...
   @NotNull private final TransformClient transformClient;
-  private final String callbackEndpoint;
+  @NotEmpty private final String callbackEndpoint;
+  @NotEmpty private final String retrieveEndpoint;
 
   public IngestServiceImpl(
       @NotNull final S3StorageAdaptor s3Adaptor,
       @NotNull final TransformClient transformClient,
-      @Value("${endpointUrl.ingest.callback}") String callbackEndpoint) {
+      @NotEmpty @Value("${endpointUrl.ingest.callback}") final String callbackEndpoint,
```
jhunzik: It doesn't have to be in this PR, but we should move the property injection out to the config file. Similar to how we do it in the S3Adaptor. This keeps our Spring stuff separated from our business logic.

multi-int-store/src/main/java/com/connexta/multiintstore/services/impl/SearchServiceImpl.java
```diff
@@ -9,23 +9,42 @@
...
 public class SearchServiceImpl implements SearchService {
 
-  private final IndexedMetadataRepository indexedMetadataRepository;
+  @NotNull private final IndexedMetadataRepository indexedMetadataRepository;
+
+  @NotEmpty private final String retrieveEndpoint;
+
+  public SearchServiceImpl(
+      @NotNull final IndexedMetadataRepository indexedMetadataRepository,
+      @NotEmpty @Value("${endpointUrl.ingest.retrieve}") final String retrieveEndpoint) {
```
jhunzik: Same as previous comment

ingest/src/main/java/com/connexta/ingest/service/impl/IngestServiceImpl.java
emmberk: :+1: Let's address that in a different PR.

emmberk:
rebased

emmberk:
`./gradlew clean build` and `./gradlew dependencyCheckAnalyze --info` succeeded locally for me

emmberk:
I manually heroed this

emmberk:
build now

emmberk:
build now

emmberk:
The CI build failed because of a known issue downloading CVEs. I'm overriding that build check because OWASP succeeded locally.

PR #55, opened 2019-06-19
Fail the build if the current JVM is not java 11

Kjames5269:
Fail the build and point them to the readme


emmberk:
build now

Kjames5269:
Failed on Owasp

jhunzik:
✅ Hero successful 🎉 Running with Java 11 works as it should and trying to build with another version will fail the build with a message to change it.

PR #59, opened 2019-06-20
Mocks RestTemplate for ingest test

vinamartin:
Mock RestTemplate for test since the RestTemplate is how we do the post to transform service. 
Assuming RestTemplate does what we think it does, this means we don't have to mock transform service.

What I'm less sure about is what meaningful things there are to assert here and what happens when errors occur on the RestTemplate postForObject.


ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
```diff
@@ -42,6 +52,32 @@ public void beforeEach() {
...
+                .contentType(MediaType.MULTIPART_FORM_DATA))
+        .andExpect(status().isOk());
+
+    verify(restTemplate, times(1))
+        .postForObject(
+            eq("https://localhost/transform"), any(HttpEntity.class), eq(TransformResponse.class));
+  }
+
+  @Test
+  public void testUnsuccessfulRequest() {}
```
jhunzik: Is this empty in anticipation for the real test?

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
```diff
@@ -42,6 +52,32 @@ public void beforeEach() {
   @Test
   public void testContextLoads() {}
 
+  @Test
+  public void testSuccessfulRequest() throws Exception {
```
jhunzik: Since this just tests if we called the transform service, should it be called `test SuccessfulTransformRequest`?

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
vinamartin: Yes, I should have probably put a TODO there.

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
vinamartin: Yeah that's probably more accurate.

vinamartin:
Looks like you can give RestTemplate an errorhandler which is how you would might mock failures. 
Should that be handled in another PR to add errorhandling?

jhunzik:
We'll be putting out PRs that handle all sorts of error handling. I would think it could go with one of those.

emmberk:
build now

emmberk:
`./gradlew clean build` and `./gradlew dependencyCheckAnalyze --info` succeeded for me locally.

PR #62, opened 2019-06-20
Updated the ingest endpoint to return 202 Accepted instead of 200 OK to comply with the API

emmberk:
https://github.com/connexta/ion-ingest-api/blob/ion-ingest-api-0.1.1/rest/specs/src/main/filtered-resources/openapi/openapi.yaml#L87-L89


emmberk:
rebased

build now

emmberk:
I heroed this

PR #63, opened 2019-06-20
Added a retrieve API

emmberk:
Both ingest and the multi-int-store services will need a retrieve endpoint. This PR is an initial step to pull out the endpoint into a common place so that it doesn't need to be duplicated. There will be future PRs to add the retrieve endpoint to the multi-int-store.


search/build.gradle
```diff
@@ -5,15 +5,3 @@
  * https://www.gnu.org/licenses/lgpl-3.0.html
  */
 /* Build Script */
-
-bootJar {
-    launchScript()
```
Kjames5269: ❓ Why are we removing the docker and boot jar from search?

emmberk: I think we had it there so that the org.springframework.boot wouldn't complain that it didn't have a main class. Now that we moved that plugin to the sub modules that are actually spring boot, I don't think we need that stuff.

emmberk: I actually just realized that there is `@SpringBootApplication` in the search app. I will update the build.gradle to match the other sub-projects.

emmberk:
rebased

emmberk:
build now

emmberk:
build now

emmberk:
I heroed this PR.

retrieve-api/src/main/java/com/connexta/retrieve/rest/spring/RetrieveApi.java
```diff
@@ -0,0 +1,21 @@
...
+import org.springframework.core.io.Resource;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+
+public interface RetrieveApi {
+
+  @GetMapping(value = "/retrieve/{id}")
+  default ResponseEntity<Resource> retrieve(@PathVariable final String id) {
```
ahoffer: I don't think we should return a ResponseEntity from an API. ResponseEntity is an implementation detail and returning from the method is a leaky abstraction. I think we should return the Resource.

emmberk: `ResponseEntity` isn't exposed beyond the endpoint. See the `RetrieveController#retrieve`. The ingest API defined by a yaml spec returns `ResponseEntity<Void>` (see https://github.com/connexta/ion-ingest-api/blob/ion-ingest-api-0.1.1/rest/specs/src/main/filtered-resources/openapi/openapi.yaml).

How would we return different status codes without using `ResponseEntity`? We also need to add a file name to the header so that the product is download with the original filename and extension. Feel free to make a new PR to fix it if you know how to fix these issues without using `ResponseEntity`.

PR #64, opened 2019-06-20
Add error handling for when S3 is down or unable to process requests

jhunzik:


ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
```diff
@@ -49,9 +55,59 @@ public void beforeEach() {
...
+                .accept(MediaType.APPLICATION_JSON)
+                .contentType(MediaType.MULTIPART_FORM_DATA))
+        .andExpect(status().is5xxServerError());
+
+    // verify
+    verifyZeroInteractions(restTemplate);
+  }
+
+  @Test
+  public void testS3UnableToStore() throws Exception{
```
jhunzik: This is super similar to the other S3 test. I'm open to suggestions.

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
```diff
@@ -49,9 +55,59 @@ public void beforeEach() {
     mvc = MockMvcBuilders.webAppContextSetup(wac).build();
   }
 
+  @After
+  public void afterEach() {
```
jhunzik: Spring doesn't manage our mock lifecycles, so we need to reset them between tests so we don't cross contaminate.

ingest/src/main/java/com/connexta/ingest/adaptors/S3StorageAdaptor.java
```diff
@@ -42,16 +45,25 @@ public void store(StoreRequest storeRequest, String key) throws IOException {
...
-            .bucket(s3BucketQuarantine)
-            .key(key)
-            .contentType(storeRequest.getMimeType())
-            .contentLength(storeRequest.getFileSize())
-            .metadata(ImmutableMap.of("filename", fileName))
-            .build(),
-        RequestBody.fromInputStream(
-            storeRequest.getFile().getInputStream(), storeRequest.getFile().getSize()));
+    try {
+      s3Client.putObject(
```
emmberk: `S3Client#putObject` can throw `AwsServiceException`, `SdkClientException`, and `S3Exception`. Should we be explicitly catch these three types of exceptions here and also `RuntimeException`? There are also many different extensions of these 3 Exception classes. that I think we'd want to handle differently, e.g. `NoSuchKeyException`, `NoSuchBucketException`, and `RetryableException`.

This all doesn't need to be addressed in this PR, but it would be worth at least catching those four exceptions; updating the message in the StorageException to match the javadoc of the `AwsServiceException`, `SdkClientException`, and `S3Exception` classes; and adding an itest for each.

ingest/src/main/java/com/connexta/ingest/exceptions/StorageException.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.ingest.exceptions;
+
+public class StorageException extends RuntimeException {
```
emmberk: ❓Why does this extend RuntimeException?

emmberk: ✏️Add the throws to the StorageAdaptor interface and add javadoc to that method.

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
emmberk: https://github.com/connexta/multi-int-store/pull/64#discussion_r296045332 I think this is right. We should have a test method for every exception that can be thrown. This is easier with parameterized tests, of course.

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
emmberk: ✏️ Maybe move this to the `@Before` method instead

ingest/src/main/java/com/connexta/ingest/exceptions/StorageException.java
```diff
@@ -0,0 +1,13 @@
...
+  public StorageException(String message, Throwable cause) {
```
Kjames5269: ✏️ This doesn't need to be addressed in this PR but it'd be neat to have this extend a `RetryableException`

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
jhunzik: I think I remember there being a possible timing issue when you reset mocks before a test.

ingest/src/main/java/com/connexta/ingest/exceptions/StorageException.java
jhunzik: I was basing it off of the S3 exceptions that were thrown. Do you know a better option?

emmberk: 🤷‍♀ That seems like a good enough reason to me.

ingest/src/main/java/com/connexta/ingest/adaptors/S3StorageAdaptor.java
```diff
@@ -42,16 +45,25 @@ public void store(StoreRequest storeRequest, String key) throws IOException {
...
+              .key(key)
+              .contentType(storeRequest.getMimeType())
+              .contentLength(storeRequest.getFileSize())
+              .metadata(ImmutableMap.of("filename", fileName))
+              .build(),
+          RequestBody.fromInputStream(
+              storeRequest.getFile().getInputStream(), storeRequest.getFile().getSize()));
+
+    } catch (SdkClientException e) {
+      throw new StorageException(
```
vinamartin: ❓ Should we do the same thing on client fail and server fail?

ingest/src/main/java/com/connexta/ingest/exceptions/StorageException.java
jhunzik: `RetryableException` is actually a final class so we wouldn't be able to have a custom class extend it.

ingest/src/main/java/com/connexta/ingest/adaptors/S3StorageAdaptor.java
jhunzik: By client and server, what are you referring to?

ingest/src/main/java/com/connexta/ingest/adaptors/S3StorageAdaptor.java
```diff
@@ -38,20 +41,31 @@ public S3StorageAdaptor(final S3Client s3Client, final String s3BucketQuarantine
...
+              storeRequest.getFile().getInputStream(), storeRequest.getFile().getSize()));
+
+    } catch (SdkServiceException e) {
+      throw new StorageException(
+          "S3 was unable to store " + key + " in bucket " + s3BucketQuarantine, e);
+    } catch (SdkClientException e) {
+      throw new StorageException(
+          "S3 was unavailable and could not store " + key + " in bucket " + s3BucketQuarantine, e);
+    } catch (RuntimeException e) {
+      throw new StorageException("Error storing " + key + " in bucket " + s3BucketQuarantine, e);
```
emmberk: ✏️ Could we get a runtime exception from the input stream stuff? Can we make it so that we only catch runtime exception thrown by the .putObject method?

jhunzik: The input stream would throw an IO Exception

emmberk: Anything could throw a RuntimeException. I think we'd only want to catch and log the RuntimeException if it comes from the S3Client:
```java
final PutObjectRequest putObjectRequest =
    PutObjectRequest.builder()
        .bucket(s3BucketQuarantine)
        .key(key)
        .contentType(storeRequest.getMimeType())
        .contentLength(storeRequest.getFileSize())
        .metadata(ImmutableMap.of("filename", fileName))
        .build();
final RequestBody requestBody =
    RequestBody.fromInputStream(
        storeRequest.getFile().getInputStream(), storeRequest.getFile().getSize());
try {
  s3Client.putObject(putObjectRequest, requestBody);
} catch (SdkServiceException e) {
  throw new StorageException(
      "S3 was unable to store " + key + " in bucket " + s3BucketQuarantine, e);
} catch (SdkClientException e) {
  throw new StorageException(
      "S3 was unavailable and could not store " + key + " in bucket " + s3BucketQuarantine, e);
} catch (RuntimeException e) {
  throw new StorageException("Error storing " + key + " in bucket " + s3BucketQuarantine, e);
}
```

ingest/src/main/java/com/connexta/ingest/adaptors/S3StorageAdaptor.java
vinamartin: client exception vs. server exception. It's possible we don't have to do anything different, just wanted to make sure

jhunzik: Unless I'm mistaken, the two exceptions we catch here should take care of it.

jhunzik:
build now

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
```diff
@@ -49,9 +55,81 @@ public void beforeEach() {
...
+                .param("fileSize", "10")
+                .param("fileName", "file")
+                .param("title", "qualityTitle")
+                .param("mimeType", "plain/text")
+                .header("Accept-Version", "1.2.1")
+                .accept(MediaType.APPLICATION_JSON)
+                .contentType(MediaType.MULTIPART_FORM_DATA))
+        .andExpect(status().is5xxServerError());
+
+    // verify
```
emmberk: ✏️Move verify comment above the perform

ingest/src/main/java/com/connexta/ingest/adaptors/S3StorageAdaptor.java
```diff
@@ -38,20 +41,32 @@ public S3StorageAdaptor(final S3Client s3Client, final String s3BucketQuarantine
...
             .metadata(ImmutableMap.of("filename", fileName))
-            .build(),
+            .build();
+    final RequestBody requestBody =
         RequestBody.fromInputStream(
-            storeRequest.getFile().getInputStream(), storeRequest.getFile().getSize()));
+            storeRequest.getFile().getInputStream(), storeRequest.getFile().getSize());
+
+    try {
+      LOGGER.info("Storing {} in bucket \"{}\" with key \"{}\"", fileName, s3BucketQuarantine, key);
```
emmberk: ✏️ Move this line out of the try catch. We don't want to catch RuntimeExceptions from the LOGGER.

emmberk:
I heroed this PR on top of https://github.com/connexta/multi-int-store/pull/66.

jhunzik:
build now

emmberk:
rebased

build now

PR #65, opened 2019-06-21
Add simple error handling for transformation service errors

vinamartin:
If any exception is thrown when communicating via the rest template, throw an error and return that back to the caller of /ingest . 
Implement retry logic at a later time.

### DIFF
--- a/ingest/src/main/java/com/connexta/ingest/config/ApplicationConfiguration.java
+++ b/ingest/src/main/java/com/connexta/ingest/config/ApplicationConfiguration.java
@@ -11,6 +11,7 @@ import org.springframework.boot.web.client.RestTemplateBuilder;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Profile;
+import org.springframework.web.client.DefaultResponseErrorHandler;
 import org.springframework.web.client.RestTemplate;
 
 @Configuration
@@ -19,6 +20,6 @@ import org.springframework.web.client.RestTemplate;
 public class ApplicationConfiguration {
   @Bean
   public RestTemplate restTemplate(RestTemplateBuilder builder) {
-    return builder.build();
+    return builder.errorHandler(new DefaultResponseErrorHandler()).build();
   }
 }
--- a/ingest/src/main/java/com/connexta/ingest/endpoint/IngestController.java
+++ b/ingest/src/main/java/com/connexta/ingest/endpoint/IngestController.java
@@ -7,6 +7,7 @@
 package com.connexta.ingest.endpoint;
 
 import com.connexta.ingest.exceptions.StorageException;
+import com.connexta.ingest.exceptions.TransformException;
 import com.connexta.ingest.rest.spring.IngestApi;
 import com.connexta.ingest.service.api.IngestService;
 import java.io.IOException;
@@ -45,7 +46,7 @@ public class IngestController implements IngestApi {
     try {
       ingestService.ingest(acceptVersion, fileSize, mimeType, file, title, fileName);
       return new ResponseEntity(HttpStatus.ACCEPTED);
-    } catch (IOException | StorageException e) {
+    } catch (IOException | StorageException | TransformException e) {
       LOGGER.warn("Unable to ingest", e);
       return new ResponseEntity(HttpStatus.INTERNAL_SERVER_ERROR);
     }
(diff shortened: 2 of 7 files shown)


ingest/src/main/java/com/connexta/ingest/config/ApplicationConfiguration.java
```diff
@@ -19,6 +20,6 @@
 public class ApplicationConfiguration {
   @Bean
   public RestTemplate restTemplate(RestTemplateBuilder builder) {
-    return builder.build();
+    return builder.errorHandler(new DefaultResponseErrorHandler()).build();
```
ahoffer: Does the default handler need to be set explicitly?

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
```diff
@@ -69,15 +77,33 @@ public void testSuccessfulTransformRequest() throws Exception {
...
   }
 
-  // TODO Test what happens when an error occurs when hitting transform service
   @Test
-  public void testUnsuccessfulRequest() {}
+  public void testUnsuccessfulRequest() throws Exception {
+    server
+        .expect(requestTo("https://localhost/transform"))
+        .andRespond(
+            withStatus(HttpStatus.BAD_REQUEST)
```
emmberk: ✏️ Should we have tests for each of the error codes that the Transform endpoint can return?
https://github.com/connexta/ion-transformation-api/blob/5f7ada923087ca6ce9334111bef59440129fd009/rest/specs/src/main/filtered-resources/openapi/openapi.yaml#L42-L56

ingest/src/main/java/com/connexta/ingest/endpoint/IngestController.java
```diff
@@ -44,7 +46,7 @@ public IngestController(IngestService ingestService) {
     try {
       ingestService.ingest(acceptVersion, fileSize, mimeType, file, title, fileName);
       return new ResponseEntity(HttpStatus.ACCEPTED);
-    } catch (IOException e) {
+    } catch (IOException | HttpClientErrorException | UnknownHttpStatusCodeException e) {
```
emmberk: ✏️ It's not clear in this class that these 2 exceptions should be caught. Can the exceptions be caught closer to where we try to hit the transform endpoint or the exceptions included in the method signatures?

ingest/src/main/java/com/connexta/ingest/config/ApplicationConfiguration.java
vinamartin: No, but I'm doing it here for clarity since we didn't realize it exists. 
Previously, I was using a custom error handler so this demonstrates how we would use one if we need to swap in the future.

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
vinamartin: I was thinking about this, but the handler throws the same exception for all of them. It's fairly trivial to add in the tests for all of them, though, so I can add them. 👍

emmberk: I'd be fine saving writing those tests for when we implement contract testing. For now, could you at least write a code comment in the tests to prevent future confusion, like
```java
\\ The error handler throws the same exception for all non-202 status codes returned by the transformation endpoint
```
?

ingest/src/main/java/com/connexta/ingest/endpoint/IngestController.java
vinamartin: I agree, it's a little weird how the error handler works since the exceptions aren't checked. 
We could catch it in the client but then we'd still have to rethrow it up so we can return the right exception to the caller.

emmberk: I suggest adding `throws HttpClientErrorException, UnknownHttpStatusCodeException` to `IngestService#ingest` or creating a custom exception like https://github.com/connexta/multi-int-store/pull/64.

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
vinamartin: Sure this works.

ingest/src/main/java/com/connexta/ingest/TransformException.java
```diff
@@ -0,0 +1,4 @@
+package com.connexta.ingest;
+
+public class TransformException extends Exception {
```
emmberk: ✏️ TransformException sounds like this is thrown when is failed to be transformed, but it really means that we weren't able to submit the transform request. I suggest renaming to IngestException or something else

emmberk:
Rebased and squashed commits

build now

emmberk:
I rebased, squashed commits, and manually tested that ingest still works.

build now

PR #66, opened 2019-06-23
Cleaned up dependencies now that not every subproject is a spring-boot app

emmberk:
https://github.com/connexta/multi-int-store/pull/63 updated the build.gradle files so that not every project must be a spring-boot app. However, some dependencies were missed in the move. This PR addresses those and also does some general clean up, including OWASP mitigations that are no longer necessary after recent version upgrades.

### DIFF
--- a/buildSrc/src/main/groovy/Dependencies.groovy
+++ b/buildSrc/src/main/groovy/Dependencies.groovy
@@ -8,31 +8,26 @@
 class Dependencies {
 
   //  Dependencies
-  static def awsBom = { it=Versions.aws -> "software.amazon.awssdk:bom:${it}" }
   static def awsS3 = { it=Versions.aws -> "software.amazon.awssdk:s3:${it}" }
-  static def commonsIO = { it=Versions.commonsIO -> "commons-io:commons-io:${it}" }
-  static def guava = { it=Versions.guava -> "com.google.guava:guava:${it}" }
-  static def ingestAPI = { it=Versions.ingestAPI -> "com.connexta.ingest:ingest-api-rest-spring-stubs:${it}" }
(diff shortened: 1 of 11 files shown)


emmberk:
I heroed this

buildSrc/src/main/groovy/Dependencies.groovy
```diff
@@ -8,31 +8,26 @@
 class Dependencies {
 
   //  Dependencies
-  static def awsBom = { it=Versions.aws -> "software.amazon.awssdk:bom:${it}" }
   static def awsS3 = { it=Versions.aws -> "software.amazon.awssdk:s3:${it}" }
+  static def lombok = { it=Versions.lombok -> "org.projectlombok:lombok:${it}" }
```
Kjames5269: ✏️ should we have alphabetical ordering?

ingest/build.gradle
```diff
@@ -10,14 +10,13 @@ plugins {
 }
 
 dependencies {
-    implementation(Dependencies.json())
+    implementation(Dependencies.springBootStarterWeb())
     implementation(Dependencies.commonsIO())
     implementation(Dependencies.ingestAPI())
     implementation(Dependencies.transformAPI())
     implementation(Dependencies.awsS3())
-    implementation platform((Dependencies.awsBom()))
```
Kjames5269: ❓ Why remove these dependencies?

emmberk: We don't need them, right?

jhunzik: We only need it if we add more AWS sdk dependencies. It ensure they are all compatible version.

emmberk: Thanks for the explanation. It is not needed right now, so I think we're good to remove it to keep our dependencies light and maintainable.

emmberk:
build now

emmberk:
build now

emmberk:
build now

PR #67, opened 2019-06-23
Implemented a retrieve endpoint for the multi-int-store service

emmberk:
The multi-int-store will be updated to use its own S3 bucket for product storage. This PR adds the common S3 configurations, storage implementation, and retrieve endpoint to the multi-int-store app.

This PR is still a draft because the itests are broken.


emmberk:
I'm closing this PR for now because we are going to implement the StorageController.

PR #70, opened 2019-06-25
Adding error handling when retrieving and connexting to Solr

Kjames5269:
More test coming soon <sup>tm</sup>


multi-int-store/src/main/java/com/connexta/multiintstore/common/StorageManager.java
```diff
@@ -66,10 +63,12 @@ private void handleCallback(MetadataCallback callback, String ingestId) {
     LOGGER.info("Metadata");
     //  TODO :: Check Markings
 
+    String contents = null;
```
emmberk: ✏️ Why was this moved out?

multi-int-store/src/main/java/com/connexta/multiintstore/common/StorageManager.java
```diff
@@ -50,9 +50,6 @@ private void sortCallbacks(Object callback, String ingestId) {
   }
 
   public void handleGeneralCallback(Object callback, String ingestId) {
```
emmberk: ✏️ Make it clear that Exceptions from this method should be caught. Same for `#handleCallback(MetadataCallback, String)`

multi-int-store/src/main/java/com/connexta/multiintstore/common/ServerException.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.multiintstore.common;
+
+public class ServerException extends RuntimeException {
```
emmberk: ✏️ Improve the name of ServerException and ClientException to be more specific to this app, like `MetadataRetrievalServerException`

multi-int-store/src/main/java/com/connexta/multiintstore/controllers/StorageController.java
```diff
@@ -36,16 +41,20 @@ public ResponseEntity store(
...
-    if (callback != null) {
+    try {
+      final Object callback = validator.parse(body);
       manager.handleGeneralCallback(callback, ingestId);
-    } else {
+    } catch (UnsupportedCallbackException e) {
+      LOGGER.info("Returning Bad Request", e);
       return new ResponseEntity(HttpStatus.BAD_REQUEST);
+    } catch (ClientException | ServerException e) {
+      LOGGER.info("Returning Internal Error due to retrieval", e);
```
emmberk: ✏️ `Returning Internal Error because the metadata retrieval failed`
and same for below

multi-int-store/src/main/java/com/connexta/multiintstore/controllers/StorageController.java
```diff
@@ -36,16 +41,20 @@ public ResponseEntity store(
       @PathVariable String ingestId, @RequestBody(required = false) JsonNode body) {
     LOGGER.info("Received callback for ingestId {}", ingestId);
 
-    Object callback = validator.parse(body);
-
-    if (callback != null) {
+    try {
+      final Object callback = validator.parse(body);
```
emmberk: ❓Could this method throw any exceptions? Where are these exceptions handled?

multi-int-store/src/main/java/com/connexta/multiintstore/common/DataRetriever.java
```diff
@@ -15,24 +15,42 @@
...
     headers.set("Accept-Version", callbackAcceptVersion);
     headers.setAccept(Collections.singletonList(MediaType.valueOf(mediaType)));
 
-    return new RestTemplate()
-        .exchange(url, HttpMethod.GET, new HttpEntity<String>(headers), clazz)
-        .getBody();
+    try {
+      return restTemplate
+          .exchange(url, HttpMethod.GET, new HttpEntity<String>(headers), clazz)
+          .getBody();
```
emmberk: ❓Could this throw a RuntimeException? Should that be caught here?

multi-int-store/src/main/java/com/connexta/multiintstore/controllers/StorageController.java
```diff
@@ -36,16 +41,20 @@ public ResponseEntity store(
...
-    } else {
+    } catch (UnsupportedCallbackException e) {
+      LOGGER.info("Returning Bad Request", e);
       return new ResponseEntity(HttpStatus.BAD_REQUEST);
+    } catch (ClientException | ServerException e) {
+      LOGGER.info("Returning Internal Error due to retrieval", e);
+      return new ResponseEntity(HttpStatus.INTERNAL_SERVER_ERROR);
+    } catch (StorageException | DuplicateIdException e) {
+      LOGGER.info("Returning Internal Error due to storage", e);
+      return new ResponseEntity(HttpStatus.INTERNAL_SERVER_ERROR);
```
emmberk: ✏️ I know that the implementation of this will probably changes, but it's weird to me that we're catching exceptions here that are only thrown for metadata callbacks. I think at least we could update the log messages to say that the problem was with the metadata.

multi-int-store/src/main/java/com/connexta/multiintstore/services/impl/IndexedMetadataDao.java
```diff
@@ -23,16 +25,34 @@ public IndexedMetadataDao(IndexedMetadataRepository repository) {
...
-    return repository.findById(id);
+    try {
+      return repository.findById(id);
+    } catch (RuntimeException e) {
+      throw new StorageException("Could not connect to Solr");
+    }
   }
 
   @Override
   public void save(IndexedProductMetadata indexedProductMetadata) {
```
emmberk: ✏️ Make it clear that Exceptions for these methods should be caught

multi-int-store/src/main/java/com/connexta/multiintstore/services/impl/IndexedMetadataDao.java
```diff
@@ -23,16 +25,34 @@ public IndexedMetadataDao(IndexedMetadataRepository repository) {
...
+    Optional<IndexedProductMetadata> document = getById(indexedProductMetadata.getId());
+    document.ifPresent(
+        (metadata) -> {
+          throw new DuplicateIdException(
+              "Metadata with id " + metadata.getId() + " already exists");
+        });
+    try {
+      repository.save(indexedProductMetadata);
+    } catch (RuntimeException e) {
+      throw new StorageException("Could not connect to Solr");
```
emmberk: ❓ Could something other than not being able to connect cause a RuntimeException?

multi-int-store/src/main/java/com/connexta/multiintstore/common/StorageManager.java
Kjames5269: I was originally doing something with this but didn't change this back after the refactor. Good catch

multi-int-store/src/main/java/com/connexta/multiintstore/controllers/StorageController.java
Kjames5269: validator.parse can throw an UnsupportedCallbackException and that should be it, the handleGeneralCallback throws all the runtime exceptions

multi-int-store/src/main/java/com/connexta/multiintstore/services/impl/IndexedMetadataDao.java
Kjames5269: Maybe, I'm unsure though. I have a hunch for our situation if this code is running it's due to a connection error. I'll make it more general

multi-int-store/src/main/java/com/connexta/multiintstore/controllers/StorageController.java
emmberk: Might be worth moving the validator.parse out of this try/catch then

multi-int-store/src/main/java/com/connexta/multiintstore/callbacks/CallbackValidator.java
```diff
@@ -37,7 +37,7 @@ public Object parse(JsonNode body) {
 
     LOGGER.warn("{} is not a valid callback", body);
 
-    return null;
+    throw new UnsupportedCallbackException(body.toString());
```
emmberk: ✏️ Should the exception contain a message like "Callback is not supported" instead of the body of the json?

multi-int-store/src/test/java/com/connexta/multiintstore/common/DataRetrieverTest.java
```diff
@@ -0,0 +1,98 @@
...
+  @After
+  public void afterEach() {
+    Mockito.reset(restTemplate);
+  }
+
+  @Test
+  public void retrieverTest() throws Exception {
+    String url = "nice.local";
+    MediaType mediaType = MediaType.APPLICATION_JSON;
+    JSONObject jsonRetval = new JSONObject().put("Quality", "Test");
```
emmberk: ✏️ Is this shorthand for `jsonRetrieval`?

Kjames5269: Yes, would you prefer that name instead?

emmberk: That's cool:+1: Just thought it might be a typo

multi-int-store/src/main/java/com/connexta/multiintstore/callbacks/CallbackValidator.java
Kjames5269: I mean it could, but it also has that in the name. I'm indifferent tbh

Kjames5269:
build now

Kjames5269:
build now

emmberk:
build now

PR #71, opened 2019-06-26
Added Minio Example and Deployment Script

garrettfreibott:
This PR add a /deployments folder as well as an example local deployment of the Multi Int Store that starts up a Minio server on port 9000, creates an ingest-quarantine bucket, and points Ingest to use that instead of an Amazon S3 store.

It also includes a deploy.sh file that will attempt to create the docker network needed and deploy everything via docker stacks. Ideally, you should be able to build with gradle, navigate to deployments, the just run `deploy.sh` to get everything up and running locally.

Example deploy script output:
deploy output

At the moment, it seems like it ingest and retrieval work. Making an API call to the ingest endpoint will successfully store the file in the Minio ingest-quarantine bucket, however it does return error. It looks like the Ingest app has trouble parsing Minio's success message for some reason. Making a call to the the ingest app at `/retrieve/<file_id>` will return the file.

I wasn't too sure where to put everything, so feel free to let me know if you think there's a better place for this.


deployments/local-minio/deploy.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+
+# Network name to setup and deploy stack to. Also update instances in docker-compose.yml file is modified.
+networkname="cdr_local-minio"
+
+# Deploy docker-compose.yml file via docker stacks
+function deploy_images () {
+    printf "Deploying Docker Images...\n"
+    docker stack deploy -c docker-compose.yml $networkname
```
jhunzik: ✏️ Really small nit. The variable here is actually setting the name of the stack and not the name of the network.

deployments/local-minio/deploy.sh
```diff
@@ -0,0 +1,56 @@
...
+}
+
+# Create attachable overlay docker network
+function create_network () {
+    printf "Creating docker '$networkname' network...\n"
+    docker network create --driver=overlay --attachable $networkname
+}
+
+# Check is network already does not exist or is not attachable.
+# Prompt to create if nonexistant or continue if network exists and is attachable.
```
jhunzik: ✏️ `nonexistent`

deployments/local-minio/docker-compose.yml
```diff
@@ -0,0 +1,122 @@
...
+#    3. Push those images.
+#    4. From the cloud environment, execute:
+#
+#        docker stack deploy -c <(REGISTRY=<registry-url> docker-compose config) cdr
+#
+#
+version: '3.7'
+services:
+  minio:
+    image: minio/minio:RELEASE.2019-06-19T18-24-42Z
```
jhunzik: ❓ Do we need this specific version? Can we use `minio/minio`?

vinamartin:
Is the local deployment testable enough with transform service?

deployments/local-minio/deploy.sh
garrettfreibott: I was just naming the stack the same as the network to keep things simple, I updated the comment on the variable at the top a bit to give a better description.

deployments/local-minio/docker-compose.yml
garrettfreibott: Nope, updated to `minio/minio` :+1:

garrettfreibott:
@vinamartin I was able to start everything by pointing the transform endpointUrl in the ingest_config to just point back at cdr-store for the time being. I can ingest files through the ingest endpoint, they get stored in the default ingest-quarantine bucket in minio, and can be retrieved via the cdr-ingest `/retrieve/` endpoint by giving it the file's ID in the URL. I haven't tried any other I/O operations or operations that use the transform service.

deployments/local-minio/docker-compose.yml
```diff
@@ -0,0 +1,122 @@
...
+    ports:
+      - target: 8080
+        published: 9040
+        protocol: tcp
+    networks:
+      - cdr_local-minio
+    deploy:
+      restart_policy:
+        condition: any
+    secrets:
```
Kjames5269: ❓do we still need the aws secrets for a local deployment?

deployments/local-minio/deploy.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
+
+# Name to give the docker network and stack. Update network instances in docker-compose.yml file if modified.
+networkname="cdr_local-minio"
+
+function wait_for_containers () {
+    waiting=`docker service ls | grep "cdr_local-minio_.*0/1 "`
```
Kjames5269: ✏️     `waiting=$(docker service ls | grep "cdr_local-minio_.*0/1 ")`

deployments/local-minio/docker-compose.yml
garrettfreibott: No, it doesn't need any secrets for a local deployment - it uses the dummy `./secrets/minio_secret.notsec` and `./secrets/minio_access.notsec` keys as the source files down on line 115 and 117. I wasn't sure if the `aws_s3_secret.sec` and `aws_s3_access.sec` filenames were hard-coded, so to be safe I kept all the target file names & variables the same.

deployments/local-minio/deploy.sh
garrettfreibott: :+1: Updated the other instances, and also replaced all the spots I was using "cdr_local-minio" with $networkname as well.

deployments/local-minio/deploy.sh
```diff
@@ -0,0 +1,64 @@
...
+        exit 1
+    fi
+}
+
+# Warning message and beginning of script
+printf "\
+====== WARNING: TEST DEPLOYMENT ======\n\
+This docker deployment will create a local minio server with default secret and key. Make sure this isn't being deployed into production.\n\n"
+
+read -p "Continue with deployment? (y/N) " -n 1 -r
```
ahoffer: I don't think we need a speedbump (y/N) here. I think the warning message alone is enough.

deployments/local-minio/deploy.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Check is network already does not exist or is not attachable.
+# Prompt to create if nonexistent or continue if network exists and is attachable.
+function check_network_configuration () {
+    printf "Checking Docker Network...\n"
+    network=`docker network ls | grep "$networkname"`
+    networktype=$(docker network inspect $networkname 2>/dev/null | grep -i "\"driver\": \"overlay\"")
+    attachable=$(docker network inspect $networkname 2>/dev/null | grep -i "\"attachable\": true")
+    if [ -z "$network" ]; then
+        read -p "Network '$networkname' does not exist, create it now? (y/N) " -n 1 -r
```
ahoffer: If the network doesn't exist, go ahead and create it. Print a message that a new network is being create. Do not prompt the user for input.

deployments/local-minio/deploy.sh
garrettfreibott: I'll rename it

deployments/local-minio/deploy.sh
garrettfreibott: I put a confirmation here because creating an external docker network on the user's machine seems a step outside of what a user would expect a deploy script to do - just bring up/down a docker-compose file.

deployments/local-minio/deploy.sh
garrettfreibott: I wanted to give the user a chance to read the warning and not immediately stand up an insecure deployment if it wasn't intended. It immediately grabs the typed characters from the buffer, so a quick double-y tap after starting the script will speed you right through the script prompts. If it's giving devs a headache though, it can be removed.

ahoffer:
build now

ahoffer:
build now

PR #76, opened 2019-07-02
Updated ingest service to use mis store endpoint

emmberk:


ingest/build.gradle
```diff
@@ -16,11 +16,10 @@ configurations {
 }
 
 dependencies {
-    implementation(Dependencies.springBootStarterWeb())
+    implementation("org.springframework.boot:spring-boot-starter-webflux:2.1.6.RELEASE")
```
emmberk: ✏️ @emmberk reminder to fix this

ingest/src/main/java/com/connexta/ingest/client/TransformClient.java
```diff
@@ -22,20 +23,22 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(TransformClient.class);
```
jhunzik: ✏️ Should we start using the Lombok annotation for the loggers from now on?

ingest/src/main/java/com/connexta/ingest/config/ApplicationConfiguration.java
```diff
@@ -18,8 +20,27 @@
 @EnableAutoConfiguration
```
jhunzik: This annotation tries to guess what beans the application needs. I think this is where Eric said we should be explicit.

ingest/src/main/java/com/connexta/ingest/config/ApplicationConfiguration.java
```diff
@@ -18,8 +20,27 @@
...
 @Profile("default")
 public class ApplicationConfiguration {
```
jhunzik: ✏️ Rename the class to describe what it configures: Ex. `ClientConfiguration`

ingest/src/main/java/com/connexta/ingest/client/StoreClient.java
```diff
@@ -0,0 +1,71 @@
...
+    this.storeEndpoint = storeEndpoint;
+    log.info("Store URL: {}", storeEndpoint);
+  }
+
+  /** @return the location of the product */
+  public URI store(
+      final Long fileSize, final String mimeType, final MultipartFile file, final String fileName)
+      throws StoreException {
+    final MultipartBodyBuilder builder = new MultipartBodyBuilder();
+    builder.part("fileSize", fileSize);
```
jhunzik: ❗️ Shouldn't these Strings be part of one JSON blob? We would add that JSON body as a separate part with the `application/json` contentType.

emmberk: Copying my answer from slack:

Good question. I'm not sure it we want to do this the same way as the ingest-api. I'll leave it to you to decide.

jhunzik: Yeah, you jogged my memory about the shortcomings of using the OpenAPI generators. We can leave this as is and I'll update the MIS controllers to work like Ingest.

ingest/src/main/java/com/connexta/ingest/client/TransformClient.java
ahoffer: My preference is to use the annotation.

ingest/src/main/java/com/connexta/ingest/config/ApplicationConfiguration.java
ahoffer: Should every bean be named? Or just where we have more than one bean of the same type?

ingest/src/main/java/com/connexta/ingest/client/StoreClient.java
ahoffer: I'm fine with leaving it this way. A multipart message is its own beast.

ingest/src/main/java/com/connexta/ingest/config/ApplicationConfiguration.java
```diff
@@ -18,8 +20,27 @@
...
+
```
ahoffer: In case I forget to ask, could someone explain how @Profile("profileName") works?

ingest/src/main/java/com/connexta/ingest/config/ApplicationConfiguration.java
emmberk: I removed `@EnableAutoConfiguration`.

emmberk: Right now, we are only naming beans with the same type.

ingest/src/main/java/com/connexta/ingest/config/ApplicationConfiguration.java
emmberk: Right now, it contains the configs for the TransformClient, StoreClient, and LoggingFilter. I think we can keep it as ApplicationConfiguration and split later if we need to.

emmberk:
build now

ingest/src/main/java/com/connexta/ingest/config/ApplicationConfiguration.java
emmberk: Eric said using `@EnableAutoConfiguration` is fine. We need it to create the `RestTemplateBuilder`.

jhunzik:
build now

emmberk:
rebased

build now

emmberk:
build now

emmberk:
I heroed this

buildSrc/src/main/groovy/Versions.groovy
```diff
@@ -18,6 +18,7 @@ class Versions {
     static String springBoot = "2.1.6.RELEASE"
     static String springData = "4.0.9.RELEASE"
     static String transformAPI = "0.0.1-SNAPSHOT"
+    static String reactiveStreams = "1.0.2"
```
emmberk: ✏️ alphabetize this

PR #77, opened 2019-07-02
Removed the search service

emmberk:
The search service isn't needed right now, and it will probably be implemented differently once we need it. Additionally, it is causing confusion with the search endpoint in the multi-int-store app.


emmberk:
build now

emmberk:
build now

emmberk:
build now

emmberk:
manually rebased to see if that fixes the build check

build now

PR #78, opened 2019-07-03
Enhanced logging

ahoffer:
This PR enhances logging for HTTP requests and responses. There are many examples of how to log HTTP requests and responses in Spring boot apps, but most implementations had limitations. I did not have the time to fully investigate the many options. I consider this PR a proof-of-concept to see what  logging approaches are helpful, and what limitations are in most need of remedy.

There are several changes, all to the `multi-int-store` application. If the changes are successful there, we can look at similar changes for the other apps in the repository.

**Changes**:
~~1. Default logging level is set to ERROR. This will cut down on the noise in the logs. ~~
~~2. Logging level for com.connexta packages set to DEBUG. This is to ensure we see relevant ~~information during the active development phase of the project.~~
3.  CommonsRequestLoggingFilter to logger added to log incoming HTTP requests, ~~including the body, up to five kilobytes~~. Does not log the body.
4. Added Spring Actuator and enabled all its endpoints. Actuator traces the last 100 HTTP messages in a JSON format. To view the trace, hit the URL that ends in `/actuator/httptrace`, e.g. `http://mis:8080/actuator/httptrace`

To test common log changes:
1. Deploy the code to the cloud. 
2. Tail the MIS logs
3. Send an HTTP request to MIS.
Expected result is to see the request in the MIS log.

To test actuator changes:
1. Assuming the changes are already deployed, visit the "actuator/httptrace" endpoint.
Expected result is to see HTTP requests AND responses. NOTE: Newest messages are added to the TOP of the trace, not the bottom.


multi-int-store/src/main/java/com/connexta/multiintstore/config/ApplicationConfiguration.java
```diff
@@ -42,4 +43,16 @@ public EndpointUrlRetrieve endpointUrlRetrieve() {
   public RestTemplate restTemplate(RestTemplateBuilder builder) {
     return builder.errorHandler(new DefaultResponseErrorHandler()).build();
   }
+
+  @Bean
```
jhunzik: ❓ I didn't see any info about request bodies or request parameters either in the Actuator endpoint or form the app logs. Is this expected?

multi-int-store/src/main/resources/application.yml
```diff
@@ -1 +1,12 @@
-callback-accept-version: 0.1.0-SNAPSHOT
\ No newline at end of file
+callback-accept-version: 0.1.0-SNAPSHOT
+logging:
+  level:
+    root: ERROR
+    org.springframework.web.filter.CommonsRequestLoggingFilter: DEBUG
+    com.connexta: DEBUG
```
jhunzik: ❗️ I no longer see logs from the com.connexta packages

multi-int-store/src/main/java/com/connexta/multiintstore/config/ApplicationConfiguration.java
```diff
@@ -42,4 +43,16 @@ public EndpointUrlRetrieve endpointUrlRetrieve() {
...
+  public CommonsRequestLoggingFilter requestLoggingFilter() {
```
jhunzik: ✏️ The request logs come out a bit messy. Is there a way to "pretty print" them?

buildSrc/src/main/groovy/Versions.groovy
```diff
@@ -15,6 +15,7 @@ class Versions {
     static String commonsIO = "2.6"
     static String ingestAPI = "0.1.1"
     static String lombok = "1.18.8"
+    static String springActuator = "2.1.6.RELEASE"
```
emmberk: ✏️ Dependencies.springActuator() can use the Versions.springBoot instead of needing to add this version here:
```groovy
static def springActuator = { it=Versions.springBoot -> "org.springframework.boot:spring-boot-starter-actuator:${it}" }
```

multi-int-store/src/main/resources/application.yml
```diff
@@ -1 +1,12 @@
...
+management:
+  endpoints:
+    web:
+      exposure:
+        include: "*"
+        exclude: "/actuator"
```
emmberk: ✏️ Add documentation about how to use the actuator

multi-int-store/src/main/java/com/connexta/multiintstore/config/ApplicationConfiguration.java
ahoffer: It is expected to see bodies in the app logs. 
You will not see bodies Actuator.
I'll look into it when I get local dev working or when I back to the office to get access to the cloud env.

multi-int-store/src/main/java/com/connexta/multiintstore/config/ApplicationConfiguration.java
ahoffer: Not easily. The log format is one message = one line. Logs are a bad solution for hierarchical data.
@oconnormi has a plan to stand up an ELK stack to parse and display logs. Once that happens, I believe we will be able to extract and pretty print the HTTP requests.

multi-int-store/src/main/resources/application.yml
ahoffer: Added a new section to the README, "Debugging" and sub section for "Spring Actuator"

multi-int-store/src/main/resources/application.yml
ahoffer: With latest changes, you should be able to see them:

`cdr_local-minio_multi-int-store.1.sfryp89wwq5k@aaron-dell    | 2019-07-06 19:21:49.656 ERROR 22 --- [nio-8080-exec-1] c.c.m.controllers.RetrieveController     : TESTING......`

multi-int-store/src/main/java/com/connexta/multiintstore/config/ApplicationConfiguration.java
ahoffer: I tried and tried to get the bodies logged. Never succeeded. The body is a stream, which you can only read once. There are caching classes to decorate the http filters that let you read the stream (up to limit) more than once. Couldn't get those to work. Could not even get the filter registered properly.

ahoffer:
I tried many approaches to get improved logging working yesterday. A lot of the approaches involve registering filters. Without Spring Boot, those are supposed to autoconfigure, but do not seem to be doing so. I've tried many examples I have found in different places on the web, but only Actuator and the commons request logging filter every worked. 

This is as good as I can do for now and I would like to get it merged.

PR #79, opened 2019-07-03
Added handling for query failures

emmberk:


multi-int-store/src/main/java/com/connexta/multiintstore/controllers/QueryController.java
```diff
@@ -34,12 +34,16 @@
   @RequestMapping(method = RequestMethod.GET, params = "q")
   @ResponseBody
   public ResponseEntity<List<URL>> searchKeyword(@RequestParam(value = "q") String keyword) {
+    if (StringUtils.isEmpty(keyword)) {
```
ahoffer: We might be able use the Bean Validation framework's annotation based validators.
I'm not immediately sure how. Here is a link to a Baeldung article that discusses them.
https://www.baeldung.com/spring-boot-bean-validation

emmberk:
rebased

emmberk:
rebased

multi-int-store/src/main/java/com/connexta/multiintstore/controllers/QueryController.java
emmberk: Good idea. I just pushed this update:
```java
@GetMapping
@ResponseBody
public ResponseEntity<List<URL>> searchKeyword(
    @RequestParam(value = "q") @NotEmpty final String keyword) {
  try {
    return new ResponseEntity<>(searchService.find(keyword), HttpStatus.OK);
  } catch (SearchException e) {
    LOGGER.warn("Unable to search for {}", keyword, e);
  }

  return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
}

@ControllerAdvice
private class ConstraintViolationExceptionHandler extends ResponseEntityExceptionHandler {

  @ExceptionHandler(value = {ConstraintViolationException.class})
  protected ResponseEntity<Object> handleConstraintViolation(
      @NotNull final ConstraintViolationException e, @NotNull final WebRequest request) {
    final String message = e.getMessage();
    final HttpStatus httpStatus = HttpStatus.BAD_REQUEST;
    LOGGER.warn("Request is invalid: {}. Returning {}.", message, httpStatus, e);
    return handleExceptionInternal(e, message, new HttpHeaders(), httpStatus, request);
  }
}
```

emmberk:
build now

PR #80, opened 2019-07-03
Implement the new storage operations for the Multi-Int-Store

jhunzik:
Just wanted to get the draft up early so we're all on the same page. I still need to clean up the tests.


emmberk:
build now

multi-int-store/src/main/java/com/connexta/multiintstore/common/StorageManager.java
```diff
@@ -6,96 +6,72 @@
...
+    final String key = UUID.randomUUID().toString().replace("-", "");
 
-  private void handleCallback(ProductCallback callback) {
-    LOGGER.info("Product");
-    //  TODO :: Check Markings
-    //  TODO :: Store Product
+    // Store in S3
+    s3.store(mimeType, file, fileSize, fileName, key);
+    // return product location
+    return new URL(retrieveEndpoint + key);
```
emmberk: ✏️ How should failing to construct the URL be handled?

multi-int-store/src/main/java/com/connexta/multiintstore/common/StorageManager.java
```diff
@@ -6,96 +6,72 @@
...
+  @NotNull private final Dao<IndexedProductMetadata, String> cstDao;
 
-  @Autowired
   public StorageManager(
-      Dao<Product, String> productDao,
-      Dao<IndexedProductMetadata, String> cstDao,
-      DataRetriever retriever) {
-    this.productDao = productDao;
+      @NotEmpty @Value("${endpointUrl.retrieve}") final String retrieveEndpoint,
+      @NotNull final S3StorageAdaptor s3,
```
emmberk: ✏️ Inject `StorageAdaptor` instead

multi-int-store/src/main/java/com/connexta/multiintstore/common/adaptors/S3StorageAdaptor.java
```diff
@@ -50,9 +48,14 @@ public void store(
...
 
-    LOGGER.info("Storing {} in bucket \"{}\" with key \"{}\"", fileName, s3BucketQuarantine, key);
+    log.info("Storing {} in bucket \"{}\" with key \"{}\"", fileName, s3BucketQuarantine, key);
     try {
       s3Client.putObject(putObjectRequest, requestBody);
+      log.info(
+          "Successfully stored \"{}\" in bucket \"{}\" with key \"{}\"",
+          fileName,
+          s3BucketQuarantine,
+          key);
```
emmberk: ✏️ Move this out of the try/catch

multi-int-store/src/main/java/com/connexta/multiintstore/common/StorageManager.java
jhunzik: Probably just the same way we do it in other places: catch a MalformedURLException

multi-int-store/src/main/java/com/connexta/multiintstore/common/adaptors/S3StorageAdaptor.java
jhunzik: To the end of the method?

emmberk: Yes. We don't want to catch runtime exceptions from Logger#info.

jhunzik: 🤦‍♂ I should have double checked the exceptions I was catching. Thanks!

multi-int-store/src/main/java/com/connexta/multiintstore/common/StorageManager.java
jhunzik: Actually, we already throw an IOException and the caller logs the message. I think we're ok leaving it as is.

ahoffer:
If the StorageAdaptor is the gateway to blob storage, how about we use blob (or even S3) in the name so that is clear how the class is different than StorageManager? Maybe "BlobStorageManager"?

jhunzik:
build now

jhunzik:
build now (not sure why it failed before)

jhunzik:
build now

jhunzik:
> If the StorageAdaptor is the gateway to blob storage, how about we use blob (or even S3) in the name so that is clear how the class is different than StorageManager? Maybe "BlobStorageManager"?

@ahoffer We kept the adaptor's name generic so that we could change the implementation to support any type of block store. Emily is working on some changes to the StorageManager. Think `ProductStorageManager` and `MetadataStorageManager`

emmberk:
I heroed this locally

jhunzik:
Overriding CI because it's failing to find the owasp dependency check artifact. It passed a full build with owasp locally.

PR #82, opened 2019-07-03
Removed unneeded classes

emmberk:
This PR deletes a lot of files that we don't need. It also combines the 2 `StorageException` classes.


emmberk:
build now

ahoffer:
I don't think this PR is going to generate a lot of discussion. @jhunzik I'm going to merge it now. If there are things that you think should be different, we'll address on a fix-forward basis.

PR #83, opened 2019-07-05
Local deployment with Minio

ahoffer:
What does this PR do?
 - Adds the ability to deploy the services locally. 
 - Storage is provided by Minio, which exposes and S3 interface.

Details:
This code is a modification of @garrettfreibott 's PR #71. Notable differences are:
  - Use compose-override file
  - Simplified directory names or flattened folder hierarchies
  - Remove any existing  local stack or network with the target name  and recreate them. This simplified the logic in the shell script deploy.sh, made the script more robust. Most importantly, it allows the value of the configs and secrets to be changed. Removing the stack also removed its immutable config and secret values and that allows new ones to be created.
  - Simplified Minio keys to make typing them into Minio's browser interface easier.
 - Changed network name to "cdr" to better mimic cloud deployment and reduce the size of the compose-override file.


ahoffer:
build now

docker-compose.yml
```diff
@@ -48,7 +48,6 @@ services:
       - target: 8080
         published: 9041
         protocol: tcp
-    restart: on-failure
```
ahoffer: Only supported for  docker-compose - not for docker stack

ahoffer:
THINGS THAT COULD BE BETTER or at least different:
1. Occasional formatting issue in output from deploy.sh
2. Destroying the network (so that it can be recreated) takes about 12 seconds. It is simpler than probing the network with inspect to see if it has the desired properties, but costs some extra time.
3. The deployment for cloud env is not like the deployment for local env. To improve consistency move cloud env specific files into a /deployments/_mycloud_ folder
4. Add a gradle task for local deployment

And finally, hacking bash scripts can be tedious. Consider scripting in a language like groovy instead.

ahoffer:
Got in contact with @garrettfreibott. He says his stuff now has a "clean" that will clear out immutable data. I'll go ahead and close this PR.

PR #85, opened 2019-07-06
Fix config files for local development

ahoffer:
The config files for local development needed updating to account for the blob storage moving from ingest to MIS. Normally I would wait for approvals before merging this, but because local dev on master is broken, I will merge this PR after a successful CI.


ahoffer:
build now

ahoffer:
build now

ahoffer:
build now

PR #86, opened 2019-07-08
Use docker override for local dev

ahoffer:
This PR reduces the size of the compose file in the /deployments/local-minio directory. It uses multiple docker files -- see https://docs.docker.com/compose/extends/ for details.

One limitation of multiple docker files is that you cannot override the network name and that variable substituion for network names is not an option without installing something like [envsubst](https://stackoverflow.com/questions/23620827/envsubst-command-not-found-on-mac-os-x-10-8).

NOTE: If you have already deployed locally using the deploy.sh script, you might need to do some clean up to make these changes work -- `docker stack rm <cdr_...>` or `docker network rm <cdr_...>`


ahoffer:
build now

deployments/local-minio/docker-override.yml
```diff
@@ -0,0 +1,29 @@
...
+    file: deployments/local-minio/secrets/minio_secret.notsec
+  aws_s3_access:
+    file: deployments/local-minio/secrets/minio_access.notsec
+
+configs:
+  # Directories relative to the master compose file
+  ingest_config:
+    file: deployments/local-minio/configs/ingest_config.yml
+  mis_config:
+    file: deployments/local-minio/configs/mis_config.ymlk
```
jhunzik: ✏️ Remove the extra `k`

ahoffer: Done

ahoffer:
Rebased. build now

PR #87, opened 2019-07-08
Update deploy.sh to be more user-friendly

garrettfreibott:
Updates to the script to make it more useful for developers.

- added deploy, redeploy, clean, and update commands. no command defaults to deploy for convenience.
- removed confirmation boxes now that commands are explicitly called

newupdate

### DIFF
--- a/deployments/local-minio/deploy.sh
+++ b/deployments/local-minio/deploy.sh
@@ -3,9 +3,38 @@
 # Name to give the docker network and stack. Update network instances in docker-compose.yml file if modified.
 name="cdr"
 
+helptext="\
+Usage: deploy.sh [command]
+Utility script for setting up a docker network and deploying a local instance of CDR into a docker swarm.
+'deploy' is the default option if no command is given.
+
+Commands:
+  d, deploy\t\tcreate the docker network and stack, stopping if one already exists
+  u, update\t\tattempts an in-place update of the deployed docker stack
+  c, clean\t\tremoves the deployed docker stack and network (destructive)
+  r, redeploy\t\tclean + deploy: completely removes the stack and network and deploys new instances (destructive)
+"
+
+# Removes the docker stack and network.
+# Docker networks usually take a few seconds to go down, so it waits to prevent redeployment issues
+function cleanup () {
+    printf "Removing $name docker stack and network...\n"
+    docker stack rm $name
+    docker network rm $name
+    waiting=$(docker network ls | grep "$name")
+    printf "\nWaiting for docker network '$name' to go down."
+    while [ ! -z "$waiting" ]; do
+        sleep 2
+        printf "."
+        waiting=$(docker network ls | grep " $name ")
+    done
+    printf "\nDone!\n"
+}
+
+# Waits for everything to start up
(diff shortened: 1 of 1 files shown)


deployments/local-minio/deploy.sh
```diff
@@ -1,64 +1,118 @@
...
 }
 
 # Deploy docker-compose.yml file via docker stacks
 function deploy_images () {
-    printf "Deploying Docker Images...\n"
-    docker stack deploy -c docker-compose.yml $networkname
+    printf "\nDeploying Docker Images...\n"
+    exists=$(docker stack ls | grep "$name")
+    if [ ! -z "$exists" ]; then
+        printf "[ERROR] '$name' stack already exists. Please run script with 'clean', 'update', 'redeploy', or 'help' for more info.\n"
```
ahoffer: :exclamation: If the stack exists, the update option will always print "... stack already exists" and exit.
Try executing "./deploy.sh" followed by "./deploy.sh update"

deployments/local-minio/deploy.sh
```diff
@@ -1,64 +1,118 @@
...
+        waiting=$(docker network ls | grep " $name ")
+    done
+    printf "\nDone!\n"
+}
+
+# Waits for everything to start up
 function wait_for_containers () {
-    waiting=$(docker service ls | grep ""$networkname"_.*0/1 ")
-    printf "\nWaiting for docker services to start."
+    waiting=$(docker service ls | grep ""$name"_.*0/1 ")
```
ahoffer: If the day comes when there is more than 1 container running for a service, this grep statement will stop working. I think this will work:
`grep $name_.*0/[[:digit:]]+"`

deployments/local-minio/deploy.sh
```diff
@@ -1,64 +1,118 @@
...
+"
+
+# Removes the docker stack and network.
+# Docker networks usually take a few seconds to go down, so it waits to prevent redeployment issues
+function cleanup () {
+    printf "Removing $name docker stack and network...\n"
+    docker stack rm $name
+    docker network rm $name
+    waiting=$(docker network ls | grep "$name")
+    printf "\nWaiting for docker network '$name' to go down."
```
ahoffer: :pencil2: There are two while loops. One to wait for network termination and one to wait for containers to come up. Define a waiting function: 

```
# Evaluate the first parameter passed to the function every periodically.
# When the expression evaluates to an empty string, return.
function wait_for_empty_results() {
  sleep 1
  while [[ -n  $(eval $1) ]]; do
    sleep 1
    printf "." >&2
  done
}
```

Then waiting for the network to be destroyed is:
    `wait_for_empty_results "docker network ls | grep $name"`

And waiting for the containers
    `wait_for_empty_results "docker service ls | grep $name_.*0/[[:digit:]]+"`

deployments/local-minio/deploy.sh
garrettfreibott: Nice catch! I'll add an update method that has a positive check for a pre-existing stack.

ahoffer:
build now

ahoffer:
build now

PR #88, opened 2019-07-08
Implement get operations in the multi int store

ahoffer:
What does this PR do?
Implements the resource fetching operations for the /mis/product and mis/product/{ID}/{metadataType} endpoints. 

What is left to do?
- ~~Possible [addition of annotation](https://github.com/connexta/multi-int-store/blob/8b999a1aa7b74620255a53790224acbfcce18789/multi-int-store/src/main/java/com/connexta/multiintstore/repositories/IndexedMetadataRepository.java#L20) to metadata repository~~ The Spring Data framework should parse the method name `findByProductIdAndMetadataType` and generate the correct Solr query. See [this](https://docs.spring.io/spring-data/solr/docs/4.0.5.RELEASE/reference/html/#solr.query-methods.criterions) and [this](https://docs.spring.io/spring-data/solr/docs/4.0.5.RELEASE/reference/html/#repositories.query-methods.query-property-expressions)
- Needs to be merged into changes with PR #80. Needs the s3 adaptor defined in that PR's version of [StorageManger](https://github.com/connexta/multi-int-store/pull/80/files#diff-336e46e38ce698e53332b8f8b85c66e6R31).
 - Tests

### DIFF
--- a/multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
+++ b/multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
@@ -8,13 +8,21 @@ package com.connexta.multiintstore.adaptors;
 
 import com.connexta.multiintstore.common.exceptions.StorageException;
 import java.io.IOException;
+import java.util.Map;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.core.io.ByteArrayResource;
+import org.springframework.http.MediaType;
 import org.springframework.stereotype.Service;
 import org.springframework.web.multipart.MultipartFile;
+import software.amazon.awssdk.core.ResponseInputStream;
 import software.amazon.awssdk.core.exception.SdkClientException;
+import software.amazon.awssdk.core.exception.SdkException;
 import software.amazon.awssdk.core.exception.SdkServiceException;
 import software.amazon.awssdk.core.sync.RequestBody;
 import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.GetObjectRequest;
+import software.amazon.awssdk.services.s3.model.GetObjectResponse;
 import software.amazon.awssdk.services.s3.model.PutObjectRequest;
 import software.amazon.awssdk.utils.ImmutableMap;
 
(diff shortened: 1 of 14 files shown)


multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
```diff
@@ -67,30 +75,29 @@ public void store(
...
-  //  }
+    try (final ResponseInputStream<GetObjectResponse> getObjectResponseResponseInputStream =
+        s3Client.getObject(
+            GetObjectRequest.builder().bucket(s3BucketQuarantine).key(key).build())) {
+      final GetObjectResponse getObjectResponse = getObjectResponseResponseInputStream.response();
+      return new RetrieveResponse(
+          MediaType.valueOf(getObjectResponse.contentType()),
+          new ByteArrayResource(getObjectResponseResponseInputStream.readAllBytes()) {
+            @Override
+            public String getFilename() {
```
ahoffer: :exclamation: : The getFilename() method is overriden to return a Filename or a key. It could be confusing to ask for a filename and not know if the method will return an actual filename or something that was not the filename. Are they cases where a called would want either the filename or the key and not care which as which?

Also, are there situations where the filename can be unknown?

multi-int-store/src/main/java/com/connexta/multiintstore/services/api/RetrieveApi.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+package com.connexta.multiintstore.services.api;
+
+import org.springframework.core.io.Resource;
+import org.springframework.http.ResponseEntity;
+import org.springframework.validation.annotation.Validated;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+
+/** TODO Replace this with the real StoreAPI */
```
ahoffer: Remove TODO comment.

multi-int-store/src/main/java/com/connexta/multiintstore/common/ProductStorageManager.java
```diff
@@ -0,0 +1,49 @@
...
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.NotNull;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Service;
+import org.springframework.web.multipart.MultipartFile;
+
+@Slf4j
+@Service
+public class ProductStorageManager {
```
ahoffer: :exclamation: I'm starting to worry we have to many layers of abstraction.
ControllerX -> StorageManagerX -> StorageAdaptorX -> RepositoryX
Is there a good case for having both StorageManagers and StorageAdaptors?

multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
```diff
@@ -67,30 +75,29 @@ public void store(
...
-  //          }
-  //        });
-  //  }
+    try (final ResponseInputStream<GetObjectResponse> getObjectResponseResponseInputStream =
+        s3Client.getObject(
+            GetObjectRequest.builder().bucket(s3BucketQuarantine).key(key).build())) {
+      final GetObjectResponse getObjectResponse = getObjectResponseResponseInputStream.response();
+      return new RetrieveResponse(
+          MediaType.valueOf(getObjectResponse.contentType()),
+          new ByteArrayResource(getObjectResponseResponseInputStream.readAllBytes()) {
```
ahoffer: :exclamation: readAllBytes read the entire stream into memory. This is fine for metadata because metadata should never be more than a few hundred K. However, the product is another matter. It could be multiple gigabytes.

multi-int-store/src/main/java/com/connexta/multiintstore/controllers/RetrieveController.java
```diff
@@ -6,32 +6,48 @@
...
+      return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+    }
+
+    log.info("Successfully retrieved id={}", productId);
+
+    final Resource resource = retrieveResponse.getResource();
+    return ResponseEntity.ok()
+        .contentType(retrieveResponse.getMediaType())
+        .header(
+            HttpHeaders.CONTENT_DISPOSITION,
```
ahoffer: Use Spring's [ContentDisposition](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/http/ContentDisposition.html) class to build the header.

multi-int-store/src/main/java/com/connexta/multiintstore/common/ProductStorageManager.java
ahoffer: My recommendation is to divide up the responsibility like this:
* Controllers are the interface between HTTP messages and model objects. For example, a controller might receive a multipart http request. It should convert (or rely on registered converters) to create a model object like Product or Metadata. 
* Controllers get models or store models by passing them to the StorageManager. The storage manager is composed of multiple StorageAdaptors, each responsible for communicating with a different persistence mechanism. There is only one StorageManager class and it is concrete. There is only one instance of StorageManager in the application.
* StorageAdaptors get models or store models by working with a Spring Data repository (like Solr), or working with a client like the S3Client. StorageAdaptor is an interface and there is a concrete subclass for each type of storage.

emmberk: I think we should address this in another PR. Feel free to create an issue.

multi-int-store/src/main/java/com/connexta/multiintstore/services/api/RetrieveApi.java
jhunzik: I put this here as a reminder to replace it with the separate API artifact that we will eventually create.

multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
emmberk: I updated RetrieveResponse to include a `fileName`. All of the products in S3 should have a fileName metadata value, so I updated S3StorageAdaptor to throw an error if the name isn't there instead of defaulting to the key.

multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
emmberk: I'll ask around to see how we want to handle this. In any case, I don't think this is a blocker for this PR. We can address in a future PR because this PR is blocking other tasks.

emmberk: I talked with Eric and pushed a new commit to address this.

emmberk:
build now

emmberk:
@jhunzik and I heroed this

PR #89, opened 2019-07-10
Moved TransformationRequest to the TransformClient

emmberk:


emmberk:
build now

emmberk:
build now

ingest/src/main/java/com/connexta/ingest/service/impl/IngestServiceImpl.java
```diff
@@ -43,11 +42,7 @@ public void ingest(
     final String location = storeClient.store(fileSize, mimeType, file, fileName).toString();
     LOGGER.info("{} has been successfully stored and can be downloaded at {}", fileName, location);
 
-    final TransformRequest transformRequest = new TransformRequest();
```
jhunzik: ❓ Should the Ingest service handle the TransformException here instead of throwing it back to the controller? (Probably doesn't need to be in this PR if we decide that's what we should do)

emmberk: How would `IngestServiceImpl` handle a `v=TransformException`? Throw an `IngestException`?

jhunzik: Yeah, I was looking at from the perspective that the Ingest controller would only worry about ingest-related exceptions.

emmberk: I agree, but I don't think it's wroth fixing right now. I would see a case where the ingest controller would want to return a different status code depending on if storage failed or only transformation failed.

PR #90, opened 2019-07-10
Injected the Accept-Version for the Transformation API

emmberk:


ingest/src/test/resources/application.yml
```diff
@@ -1,4 +1,7 @@
 # Todo: Update these placeholders when we add tests that use them.
 endpointUrl:
   store: http://localhost:1232/store/
-  transform: http://localhost:1231/transform/
\ No newline at end of file
+  transform: http://localhost:1231/transform/
+endpoints:
+  transform:
+    version: testTransformApiVersion
```
emmberk: ❓ I'm not sure if we care about verifying the accept-version in the itests. I wasn't able to figure out how to inject the value into the itest class.

emmberk:
build now

ingest/src/test/resources/application.yml
jhunzik: It can be injected the same way we do it in the application code. It will just pull the value from the test `application.yml`

jhunzik: https://github.com/lessarderic/multi-int-store/commit/35291f97854d8ab06d24908fc93a30bf39dbe2e5#diff-8ffc41e3b5fbbe77437d1030547fb1dbR54
Then we can just autowire it in the test class.

emmberk: Neat. I didn't realize `@Value` worked in the itest class.

The `processIntegrationTestResources` didn't immediately work for me. I don't think it really matters that we inject that value into the test resources as well because any value in the application.yml file should work.

emmberk:
build now

PR #92, opened 2019-07-10
Added min/max validation constrains to fileSize

emmberk:


ingest/src/main/java/com/connexta/ingest/client/StoreClient.java
```diff
@@ -39,7 +41,7 @@ public StoreClient(
 
   /** @return the location of the product */
   public URI store(
-      final Long fileSize,
+      @NotNull @Min(1L) @Max(10737418240L) final Long fileSize,
```
emmberk: `@NotNull` is included because
> `null` elements are considered valid.

for [`@Min`](https://javaee.github.io/javaee-spec/javadocs/javax/validation/constraints/Min.html) and [`@Max`](https://javaee.github.io/javaee-spec/javadocs/javax/validation/constraints/Max.html).

ingest/src/main/java/com/connexta/ingest/endpoint/IngestController.java
```diff
@@ -41,10 +41,12 @@ public IngestController(IngestService ingestService) {
       MultipartFile file,
       String title,
       String fileName) {
+    // TODO validate fileSize
```
emmberk: OpenApi does not generate validator annotations for the REST API interface. It is not possible to add the annotations to the overriden method, so we should validate within the method. This will be handled in a future PR.

jhunzik:
build now

emmberk:
rebased

build now

emmberk:
I heroed this

PR #94, opened 2019-07-10
Contain MultipartFile to only the REST endpoints

emmberk:
This PR updates both the ingest app and mis app to only use MultipartFile in the `@RestController` classes to make the apps less coupled to spring web.

### DIFF
--- a/ingest/src/main/java/com/connexta/ingest/client/StoreClient.java
+++ b/ingest/src/main/java/com/connexta/ingest/client/StoreClient.java
@@ -7,6 +7,7 @@
 package com.connexta.ingest.client;
 
 import com.connexta.ingest.exceptions.StoreException;
+import java.io.InputStream;
 import java.net.URI;
 import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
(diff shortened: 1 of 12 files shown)


emmberk:
I heroed this.

emmberk:
rebased

build now

ingest/src/main/java/com/connexta/ingest/client/StoreClient.java
```diff
@@ -43,13 +44,26 @@ public StoreClient(
...
+      @NotNull final InputStream inputStream,
       @NotEmpty final String fileName)
       throws StoreException {
     final MultipartBodyBuilder builder = new MultipartBodyBuilder();
     builder.part("fileSize", fileSize);
     builder.part("mimeType", mimeType);
-    builder.part("file", file.getResource());
+    builder.part(
+        "file",
+        new InputStreamResource(inputStream) {
```
jhunzik: I wish there was a nicer way to do this

emmberk: I agree. I didn't research it much, so maybe I missed something.

PR #95, opened 2019-07-10
Move itest into their own module and create a separate build phase for them

jhunzik:


README.md
```diff
@@ -35,8 +35,8 @@ For more information about spotless checks see
 [here](https://github.com/diffplug/spotless/tree/master/plugin-gradle#custom-rules).
 
 #### Tests
-* Tests are run automatically with `./gradlew build`.
-* To skip all tests, add `-x test`.
+* All tests are run automatically with `./gradlew build`.
+* To skip all tests, add `-x test` and `-x integrationTest`.
```
emmberk: ✏️ Does `-x test` skip unit tests and `-x integrationTest` skip itests? I recommend including the command for skipping all tests and the command for just skipping itests.

ingest/build.gradle
```diff
@@ -9,10 +9,23 @@ plugins {
...
 }
 
+sourceSets {
+    integrationTest {
+        java {
+            compileClasspath += main.output + test.output
+            runtimeClasspath += main.output + test.output
+            srcDir file("src/itest/java")
+        }
+        resources.srcDir file("src/itest/resources")
```
emmberk: ✏️ consistently use `itest` or `integrationTest`
It looks like this in IntelliJ right now:
image

ingest/build.gradle
```diff
@@ -21,7 +34,14 @@ dependencies {
...
     implementation(Dependencies.ingestAPI())
     implementation(Dependencies.transformAPI())
+
     testImplementation(Dependencies.springBootStarterTest())
+
+    integrationTestImplementation(Dependencies.ingestAPI())
+    integrationTestImplementation(Dependencies.transformAPI())
+    integrationTestImplementation(Dependencies.springBootStarterTest())
+    integrationTestImplementation(Dependencies.springBootStarterWeb())
+    integrationTestImplementation(Dependencies.reactiveStreams())
```
emmberk: ❓Shouldn't the itests inherit all of the dependencies from `implementation`?

ingest/build.gradle
jhunzik: Both the module name and directory are named `itest`, not sure why IntelliJ is doing that. I don't have an issue with the longer name.

emmberk: I prefer itest

README.md
jhunzik: Ah yeah. I clarified that and must have remove it in the readme. I think i can just have test depend on integrationTest so we get:
all tests: ./gradlew build
no itests: ./gradlew build -x test

ingest/build.gradle
jhunzik: That's what I thought, but it seems like it's not the case

emmberk: I think we should figure out how to make that work. I could see an issue where we don't maintain the dependencies between implementation and itestImplementation, which means that we're running the itests with different dependencies.

jhunzik: This was a nightmare to look into, I think it should be a separate pr.

emmberk: I'm not okay merging until we figure this out. I think this change introduces a lot of risk.

emmberk: CC @lessarderic

build.gradle
```diff
@@ -86,6 +87,14 @@ subprojects {
         }
     }
 
+    idea{
```
jhunzik: IntelliJ doesn't play nice with Gradle when it comes to custom sourceSets. The best I could come up with is using the idea plugin. If this isn't worth adding another plugin, I can remove it.

emmberk:
😢

PR #97, opened 2019-07-10
Cleaned up String configs in ingest

emmberk:
Using `@Value` is preferred over `@Qualifier` in the constructor of `@Service` classes.

The mis configs will be handled in https://github.com/connexta/multi-int-store/pull/96.


emmberk:
build now

PR #101, opened 2019-07-12
Remove @Service and manually create beans in the configuration files

jhunzik:


ingest/src/main/java/com/connexta/ingest/config/ApplicationConfiguration.java
```diff
@@ -43,4 +44,18 @@ public CommonsRequestLoggingFilter requestLoggingFilter() {
   public IngestService ingestService(StoreClient storeClient, TransformClient transformClient) {
     return new IngestServiceImpl(storeClient, transformClient);
   }
+
+  @Bean
+  public StoreClient storeClient(
+      RestTemplate restTemplate, @Value("${endpointUrl.store}") String storeEndpoint) {
```
emmberk: ✏️ Add the validators (`@NotNull` and `@NotEmpty`) back to all of the constructor args

multi-int-store/src/main/java/com/connexta/multiintstore/config/ApplicationConfiguration.java
```diff
@@ -32,4 +42,27 @@ public CommonsRequestLoggingFilter requestLoggingFilter() {
...
+  @Bean
+  public SearchService searchService(
+      IndexedMetadataRepository indexedMetadataRepository,
+      @Value("${endpointUrl.retrieve}") String retrieveEndpoint) {
+    return new SearchServiceImpl(indexedMetadataRepository, retrieveEndpoint);
+  }
+
+  @Bean
+  public IndexedMetadataDao indexedMetadataDao(IndexedMetadataRepository repository) {
+    return new IndexedMetadataDao(repository);
```
emmberk: ✏️ Should we have interfaces and impls for `IndexedMetadataDao`, `ProductStorageManager`, and `MetadataStorageManager`? We can probably do this in another PR.

multi-int-store/src/test/java/com/connexta/multiintstore/TestUtilConfig.java
```diff
@@ -23,6 +23,13 @@ public S3Client s3Client() {
     return mock(S3Client.class);
   }
 
+  //  @Bean
+  //  @Profile("test")
+  //  public S3StorageAdaptor s3StorageAdaptor() {
+  //    return new S3StorageAdaptor(mock(S3Client.class), "http://tests3");
+  //    // return new S3StorageAdaptor(s3Client(), "http://tests3");
+  //  }
```
emmberk: ✏️ Should this be removed?

multi-int-store/src/test/resources/application.yml
```diff
@@ -3,4 +3,5 @@ endpointUrl:
 aws:
   s3:
     bucket:
-      quarantine: quarantine
\ No newline at end of file
+      quarantine: quarantine
+    endpointUrl: http://localhost:1234/s3
```
emmberk: ❓Why is `aws.s3.endpointUrl` needed but not `aws.s3.region`?

multi-int-store/src/main/java/com/connexta/multiintstore/config/ApplicationConfiguration.java
jhunzik: I would say we should wait until we need an interface. I'm not sure what practical value we would be adding by making an interface for these.

multi-int-store/src/test/resources/application.yml
jhunzik: I'm actually surprised that the App Context was able to be configured without this. For clarification, you typically only include the the endpointUrl if you want to override the location of S3. (For example, the Mino S3 that we use for testing)

emmberk: Both of those values are in the same constructor. I'm confused why you only need one.

jhunzik: Actually we don't need either, I forgot that we mock the S3Client in the tests. We need the bucket name for property injection when the S3Adaptor bean gets created.

ahoffer:
build now

jhunzik:
build now

jhunzik:
spotless. build now

jhunzik:
Hero'd and confirmed that all the beans were able to start and the endpoints worked as intended.

PR #102, opened 2019-07-13
Actuator for ingest and MIS services

ahoffer:
Adds in dependencies and configurations for Spring Boot Actuator. 
Configured to expose all Actuator endpoints.

To test (using *local deployment*), enter an actuator endpoint into a browser. E.g.

`http://127.0.0.1:9040/actuator/httptrace`
`http://127.0.0.1:9041/actuator`

### DIFF
--- a/buildSrc/src/main/groovy/Dependencies.groovy
+++ b/buildSrc/src/main/groovy/Dependencies.groovy
@@ -13,10 +13,10 @@ class Dependencies {
   static def ingestAPI = { it=Versions.ingestAPI -> "com.connexta.ingest:ingest-api-rest-spring-stubs:${it}" }
   static def lombok = { it=Versions.lombok -> "org.projectlombok:lombok:${it}" }
   static def reactiveStreams = { it=Versions.reactiveStreams -> "org.reactivestreams:reactive-streams:${it}" }
+  static def springActuator = { it=Versions.springBoot -> "org.springframework.boot:spring-boot-starter-actuator:${it}" }
   static def springBootStarterWeb = { it=Versions.springBoot -> "org.springframework.boot:spring-boot-starter-web:${it}" }
   static def springDataSolr = { it=Versions.springData -> "org.springframework.data:spring-data-solr:${it}" }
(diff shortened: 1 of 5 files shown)


ahoffer:
build now

buildSrc/src/main/groovy/Dependencies.groovy
```diff
@@ -16,6 +16,8 @@ class Dependencies {
   static def springBootStarterWeb = { it=Versions.springBoot -> "org.springframework.boot:spring-boot-starter-web:${it}" }
   static def springDataSolr = { it=Versions.springData -> "org.springframework.data:spring-data-solr:${it}" }
   static def transformAPI = { it=Versions.transformAPI -> "com.connexta.transformation:transformation-api-rest-spring-stubs:${it}" }
+  static def springActuator = { it=Versions.springBoot -> "org.springframework.boot:spring-boot-starter-actuator:${it}" }
```
jhunzik: ❓ I wonder if we need to lock this endpoint now at this point.

multi-int-store/src/main/resources/application.yml
```diff
@@ -8,7 +8,6 @@ management:
     web:
       exposure:
         include: "*"
```
jhunzik: I wonder if we should include all the endpoints or maybe just httptrace, health and loggers.

buildSrc/src/main/groovy/Dependencies.groovy
ahoffer: Eventually, yes. It is not something you want exposed in a production system. We can do that Until we integrate with the security services.

multi-int-store/src/main/resources/application.yml
ahoffer: I don't know which endpoints will prove useful yet. I would like to leave them open and play around with them a bit.

ahoffer:
build now

ahoffer:
build now

ahoffer:
build now

PR #105, opened 2019-07-16
Created a config file for the transform service

jhunzik:


jhunzik:
build now

jhunzik:
I hero'd this and verified that all the services with configurations started up and I was able to ingest and store a product.

PR #107, opened 2019-07-17
Increase request size limit

jhunzik:
Updates the request size limits to enable larger file uploads. The default values for file and request size are 1MB and 100Mb respectively.

To Hero: Verify you can ingest a files up to 5GB in size and confirm that it was stored in S3.


jhunzik:
build now

ahoffer:
build now

PR #108, opened 2019-07-18
Switch to AWS SDK V1 to support large file uploads to S3

jhunzik:
This PR reverts to V1 of the Amazon SDK for S3 in order to support ingest of files larger than 5GB. 

In order to test this PR, one should try to ingest a file below 5GB, a file greater than 5GB but less than 10GB, and a file greater than 10GB. The first two should ingest, albeit slowly. Files greater than 10GB should throw a validation exception in the StoreController.

UPDATE: Files larger than ~1GB will probably cause an OOM Exception. We have a ticket to fix this, but in the meantime, anything over 1MB should be good enough to test.

### DIFF
--- a/multi-int-store/src/main/java/com/connexta/multiintstore/common/ProductStorageManager.java
+++ b/multi-int-store/src/main/java/com/connexta/multiintstore/common/ProductStorageManager.java
@@ -51,7 +51,6 @@ public class ProductStorageManager {
    * RetrieveResponse}.
    */
   public RetrieveResponse retrieveProduct(@NotEmpty String id) throws StorageException {
-    // return storageAdaptor.retrieve(id);
-    return null;
+    return storageAdaptor.retrieve(id);
   }
 }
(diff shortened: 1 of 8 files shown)


multi-int-store/src/main/java/com/connexta/multiintstore/common/ProductStorageManager.java
```diff
@@ -51,6 +51,7 @@ public URI storeProduct(
    * RetrieveResponse}.
    */
   public RetrieveResponse retrieveProduct(@NotEmpty String id) throws StorageException {
-    return storageAdaptor.retrieve(id);
+    // return storageAdaptor.retrieve(id);
+    return null;
```
ahoffer: :question: Should this be returning null?

ahoffer:
I wonder if we should have just made a new AmazonAdaptor class so we could leave the S3Adaptor around for the day when we really do switch over to V2.

multi-int-store/src/main/java/com/connexta/multiintstore/common/ProductStorageManager.java
jhunzik: No, no it should not. Thanks!

jhunzik:
> I wonder if we should have just made a new AmazonAdaptor class so we could leave the S3Adaptor around for the day when we really do switch over to V2.

@ahoffer That's a good point, my only worry is that there is no timeline for when Amazon will be releasing a version of V2 with the TransferManager. That extra adaptor may stick around for a while.

jhunzik:
build now

PR #110, opened 2019-07-20
Swagger UI for ingest endpoint

ahoffer:
Enabled Swagger UI for the ingest endpoint. (It is already enabled for MIS endpoint)


ahoffer:
build now

PR #111, opened 2019-07-24
Enable local JVM monitoring

ahoffer:
Expose JMX service for Ingest and MIS. This permits monitoring the JVM including the size of the head. Also, enable and expose Java debugger.

To Test JMX:
1. Build the branch and deploy locally
2. Download and run visualvm (https://visualvm.github.io/)
3. In visualvm, add a JMX connection for ingest (localhost:10040). There should be no connection error.
4. In visualvm, add JMX connection for MIS (localhost:10041). There should be no connection error.

To test debugger. 
1. Run the app locally.
2. Attach debugger to ingest application. This is the IntelliJ configuration:

image

When the day comes to harden the system, this functionality will be turned off.

### DIFF
--- a/deployments/local-minio/README.md
+++ b/deployments/local-minio/README.md
@@ -11,12 +11,18 @@ includes a deploy.sh file that will attempt to create the docker network needed.
 * Linux/Mac
 
 ## How to Run
-From this directory, run and follow the prompts to complete deployment:
-```
+From this directory:
+```bash
 ./deploy.sh
 ```
 The script should provide appropriate error messages to resolve any issues that may occur while setting up
 the docker network or deploying the images.
+
+## How to Stop
+From this directory:
+```bash
+./deploy.sh clean
+```
 
 ## Minio
 The minio web client is exposed on port 9000 with the default login information:
--- a/deployments/local-minio/docker-override.yml
+++ b/deployments/local-minio/docker-override.yml
@@ -13,6 +13,17 @@ services:
     networks:
       - cdr
     command: -c "mkdir /data/ingest-quarantine && /usr/bin/minio server /data"
+  # Enable JVM monitoring
+  ingest:
+    ports:
+      - target: 10040
+        published: 10040
+        protocol: tcp
+  multi-int-store:
+    ports:
+      - target: 10041
+        published: 10041
+        protocol: tcp
 
 secrets:
   # Directories are relative to the master compose file
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,9 +31,6 @@ services:
       - target: 8080
         published: 9040
         protocol: tcp
-      - target: 10040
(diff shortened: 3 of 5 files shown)


ahoffer:
build now.

ahoffer:
build now now

ahoffer:
build now

clockard:
Are you enabling/exposing jmx for testing purposes or do you think this will be enabled in production as well? I know jmx opens up some security vulnerabilities so if we plan on leaving it enabled the threat model would need to be updated to address those.

ahoffer:
> Are you enabling/exposing jmx for testing purposes or do you think this will be enabled in production as well? I know jmx opens up some security vulnerabilities so if we plan on leaving it enabled the threat model would need to be updated to address those.

I thought long and hard about the implications of merging this into the code base. JMX and debugging are being enabled ONLY for development purposes.  Debugging/monitoring always have a trade-off with security.  These capabilities open a system up and expose additional attack vectors. Debugging and JMX should not be deployed in production, unless properly hardened. 

These tools are valuable and I do need them for development. I considered different ways of enabled the functionality:

1. Add env var changes to docker override file for local dev. They would only be enabled for local development and not cloud development. For cloud deployment, create another docker override file use it only when debugging in the cloud is needed.
  - Pros: dev options not deployed to cloud unless needed
  - Cons: 
    - More files, more options and therefore more complexity
    - More initial development time to get it all working correctly
    - Don't always know when you will need to connect a deubgger -- time lost rebuilding, redeploying, and repeating steps to get to the break point for the cloud.
2. Use a git patch and apply it before it as needed.
  - Pros: Super simple. No changes to controlled source files.
  - Cons: 
    - Patch quickly becomes outdated as the product evolves. 
    - Leaves dirty files in the working tree that should not be added or committed to the git index.
3. Production is months in the future. Maybe further. Enable debugging for local and cloud development. Worry about hardening the system for production when the operational date is closer.
  - Pros: 
    - Debugging/monitoring is always available
    - No special action needed to enable dev tools.
 - Cons:
  - Need to be disable before system becomes operational.

In the end, I settled on option #3. We are going to be developing for a long time still. I felt the convenience of always having them at hand outweighed the risk of forgetting about them. Even if someone forgets, there are port scans, security audits and other controls to catch this kind of thing before a system becomes operational.

The only option I completely ruled out was #2, the patch file.

@emmberk @clockard 
How do you feel about it? Would you prefer another option?

clockard:
I am fine with option 3 but would recommend creating a ticket in the backlog to turn jmx and debug off so we don't forget.

paouelle:
Aren't all our services going to sit behind a firewall which protects them and only expose the ports/protocols/apis we need to? If so then there shouldn't be any problems or impact to leave these open.

emmberk:
build now

PR #113, opened 2019-07-25
Updated the search endpoint to return URIs instead of URLs

emmberk:
This PR updates the query endpoint to return URIs instead of URLs to match the type of the `Location` header in the store product response.


emmberk:
rebased

build now

ahoffer:
Heroed in the cloud. Could send ingest. Could successfully store and retrieve a product.

ahoffer:
Tested the query endpoint in the cloud and it worked as expected.

PR #114, opened 2019-07-25
Improved mis itests

emmberk:
This PR should only be merged after https://github.com/connexta/multi-int-store/pull/113.


multi-int-store/src/main/java/com/connexta/multiintstore/config/S3StorageConfiguration.java
```diff
@@ -33,42 +33,44 @@
...
+            new EndpointConfiguration(configuration.getEndpoint(), configuration.getRegion()))
+        .withCredentials(
+            new AWSStaticCredentialsProvider(
+                new BasicAWSCredentials(
+                    configuration.getAccessKey(), configuration.getSecretKey())))
+        .enablePathStyleAccess()
+        .build();
+  }
+
+  // TODO does this need to be a separate bean?
```
jhunzik: No, we should be able to initialize it in the StorageAdaptor bean since all Spring beans are singletons by default.

multi-int-store/src/test/java/com/connexta/multiintstore/S3Tests.java
```diff
@@ -0,0 +1,385 @@
...
+  @Test
+  public void testStoreProductWhenS3IsEmpty() throws Exception {
+    // given
+    final InputStream inputStream =
+        IOUtils.toInputStream(
+            "All the color had been leached from Winterfell until only grey and white remained",
+            StandardCharsets.UTF_8);
+    final long fileSize = (long) inputStream.available();
+    final String fileName = "test_file_name.txt";
+    final MultipartBodyBuilder multipartBodyBuilder = new MultipartBodyBuilder();
```
jhunzik: We should be using a `MulitValueMap<String, Object>` to create the multipart request. https://github.com/spring-projects/spring-boot/issues/12579#issuecomment-375305699. With this change we won't need to pull in ReactiveStreams as a dependency.

The next Spring release will update the docs to make this recommendation.

multi-int-store/src/test/java/com/connexta/multiintstore/SolrTests.java
```diff
@@ -0,0 +1,1095 @@
...
+        "All the color had been leached from "
+            + queryKeyword
+            + " until only grey and white remained";
+    final String productEncoding = "UTF-8";
+    final InputStream productInputStream = IOUtils.toInputStream(productContents, productEncoding);
+    final long productFileSize = (long) productInputStream.available();
+    final String productFileName = "test_file_name.txt";
+    final String productMimeType = "text/plain";
+    final MultipartBodyBuilder storeProductRequestBodyBuilder = new MultipartBodyBuilder();
+    storeProductRequestBodyBuilder.part("fileSize", String.valueOf(productFileSize));
```
jhunzik: Same as my other comment in `S3Tests`

multi-int-store/src/main/java/com/connexta/multiintstore/config/S3StorageConfiguration.java
emmberk: Let's address this in a future PR. I'll leave the TODO comment in there.

emmberk:
I rebased, addressed comments, and fixed an issue with deploying.

build now

ahoffer:
Heroed commit 469e5a354dcdf8fb901517cfba53e6a1e47ba5c9.
Integration tests passed.
Store and Retrieve in the cloud passed.
Ingest in the cloud is behaving as expected (failured because transform service had an error.)

PR #115, opened 2019-07-29
Resolve heap overflow

ahoffer:
This PR resolves an out-of-memory condition caused by the default settings of the Spring RestTemplate. The RestTemplate buffers outgoing bytes by default. 

This PR create a new class that sends disables that behavior before sending a message. 

This PR builds on the debugging tools enhancements in branch ahoffer:jvm-monitoring

This PR fails several integration tests. Those errors might not be related to this PR. Here is one of them:

> MockHttpServletResponse:
>            Status = 400
>     Error message = Required String parameter 'fileName' is not present
>           Headers = []
>      Content type = null
>              Body = 
>     Forwarded URL = null
>    Redirected URL = null
>           Cookies = []
>   com.connexta.ingest.IngestApplicationIntegrationTest testStoreRequestServerError: FAILURE


multi-int-store/Dockerfile
```diff
@@ -1,7 +1,17 @@
-FROM openjdk:11-jre-slim
+FROM openjdk:11
```
ahoffer: The image openjdk:11 provides better debugging tools than jre-slim. Before deployment to prod, consider changing back to a slim image.

multi-int-store/src/main/java/com/connexta/multiintstore/controllers/StoreController.java
```diff
@@ -88,7 +88,7 @@ public StoreController(
       return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build();
     }
 
-    return ResponseEntity.created(location).build();
+    return ResponseEntity.created(location).body(location);
```
ahoffer: Return the URI as a header and as a the body. Spring rest controllers will serialize the body back into a URI automatically.

ingest/src/main/java/com/connexta/ingest/client/StoreClient.java
```diff
@@ -69,12 +68,7 @@ public String getFilename() {
...
     log.info("Sending POST request to {}: {}", storeEndpoint, request);
 
-    final URI location;
-    try {
-      location = restTemplate.postForLocation(storeEndpoint, request);
-    } catch (final RestClientException e) {
-      throw new StoreException("Unable to POST to store endpoint", e);
-    }
-    return location;
+    return new StreamingRestTemplate(storeEndpoint, HttpMethod.POST)
```
emmberk: ❓ Can we share the same RestTemplate instead of instantiate a new one for every store request?

ingest/src/main/java/com/connexta/ingest/client/StoreClient.java
```diff
@@ -69,12 +68,7 @@ public String getFilename() {
...
+        .exchange(request, new ParameterizedTypeReference<URI>() {});
```
emmberk: ❗️Can this `exchange` method fail? How/where do we handle failures?

ingest/src/main/java/com/connexta/ingest/common/StreamingRestTemplate.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.springframework.core.ParameterizedTypeReference;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.ResponseEntity;
+import org.springframework.http.client.SimpleClientHttpRequestFactory;
+import org.springframework.web.client.RestClientException;
+import org.springframework.web.client.RestTemplate;
+
+public class StreamingRestTemplate {
```
emmberk: ❓ Should the `RestTemplate` `Bean` in `ApplicationConfiguration` be removed? Should we use the same `StreamingRestTemplate` for sending transform requests as well?

multi-int-store/src/main/java/com/connexta/multiintstore/controllers/StoreController.java
emmberk: ❗️Why was this change needed? The location is already returned as a header, which I believe is the recommended way to handle POST requests.

deployments/local-minio/docker-override.yml
```diff
@@ -10,8 +10,11 @@ services:
     environment:
       - MINIO_ACCESS_KEY=MINIOEXAMPLEACCESSKEY
       - MINIO_SECRET_KEY=MINIOEXAMPLESECRETKEY
+      - MINIO_DOMAIN=minio
     networks:
-      - cdr
+      cdr:
+        aliases:
+          - ingest-quarantine.minio
```
emmberk: ❓Why was this needed?

ingest/Dockerfile
```diff
@@ -4,4 +4,14 @@ LABEL com.connexta.application.name=cdr-ingest
 ARG JAR_FILE
 COPY ${JAR_FILE} /cdr-ingest
 ENTRYPOINT ["/cdr-ingest"]
-EXPOSE 8080
+EXPOSE 8080 10040 10050
+# Enable JMX so the can JVM be monitored
```
emmberk: ❓Aren't these changes in another PR: https://github.com/connexta/multi-int-store/pull/111? I recommend keeping the changes in separate PRs. Does https://github.com/connexta/multi-int-store/pull/111 need to be merged first?

ahoffer:
PR #116 supercedes this one.

PR #116, opened 2019-07-29
Resolve ingest memory usage

ahoffer:
Disable HTTP buffering when storing a file.
The changes are isolated to a new bean ("nonBuggeringRestTemplate")
This might be a change we want to make to the general RestTemplate bean because it cuts down on memory usage.


ahoffer:
Missing a commit

PR #117, opened 2019-07-29
Non-buffering Rest Template

ahoffer:
Disable HTTP buffering when storing a file.
The changes are isolated to a new bean ("nonBuggeringRestTemplate")
This might be a change we want to make to the shared RestTemplate bean -- the change reduces memory usage.

### DIFF
--- a/ingest/src/main/java/com/connexta/ingest/config/StoreClientConfiguration.java
+++ b/ingest/src/main/java/com/connexta/ingest/config/StoreClientConfiguration.java
@@ -9,16 +9,26 @@ package com.connexta.ingest.config;
 import com.connexta.ingest.client.StoreClient;
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.http.client.SimpleClientHttpRequestFactory;
 import org.springframework.web.client.RestTemplate;
 
 @Configuration
 public class StoreClientConfiguration {
+
+  @Bean("nonBufferingRestTemplate")
+  public RestTemplate nonBufferingRestTemplate() {
+    SimpleClientHttpRequestFactory requestFactory = new SimpleClientHttpRequestFactory();
+    requestFactory.setBufferRequestBody(false);
+    return new RestTemplate(requestFactory);
+  }
+
   @Bean
   public StoreClient storeClient(
-      @NotNull RestTemplate restTemplate,
+      @NotNull @Qualifier("nonBufferingRestTemplate") RestTemplate restTemplate,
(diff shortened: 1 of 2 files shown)


ahoffer:
build now

ingest/src/main/java/com/connexta/ingest/config/StoreClientConfiguration.java
```diff
@@ -9,16 +9,26 @@
...
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.http.client.SimpleClientHttpRequestFactory;
 import org.springframework.web.client.RestTemplate;
 
 @Configuration
 public class StoreClientConfiguration {
+
+  @Bean("nonBufferingRestTemplate")
```
emmberk: ❓Not a blocker for this PR, but is there a guarantee that the `TransformClient` `Bean` won't pick up this `Bean`?

ahoffer: I haven't read any guarantees. I suspect the transform client would continue to pick up the unnamed RestTemplate bean, unless that bean is unavailable. In that case it would use the named RestTemplate bean.

ahoffer:
build now

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
```diff
@@ -60,17 +62,27 @@
...
+  @Qualifier("nonBufferingRestTemplate")
+  private RestTemplate nonBufferingRestTemplate;
+
+  private MockRestServiceServer storeServer;
+  private MockRestServiceServer transformServer;
 
   @Before
   public void beforeEach() {
-    server = MockRestServiceServer.createServer(restTemplate);
+    MockRestServiceServerBuilder builder = MockRestServiceServer.bindTo(nonBufferingRestTemplate);
```
ahoffer: Client and server need to use the same bean for MockServer to work. MockMvc intercepts the HTTP request send to the RestTemplate and redirects it to the mock server.

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
```diff
@@ -60,17 +62,27 @@
...
+    builder.bufferContent();
+    storeServer = builder.build();
+    transformServer = MockRestServiceServer.createServer(restTemplate);
```
emmberk: ❓Why can't you create the `storeServer` the same way as the `transformServer`?

```java
@Before
public void beforeEach() {
  storeServer = MockRestServiceServer.createServer(nonBufferingRestTemplate);
  transformServer = MockRestServiceServer.createServer(restTemplate);
}
```

ahoffer: That might be left over from something I tried originally. Let me take it out and see it the tests still pass.

ahoffer:
build now

PR #119, opened 2019-07-30
Fix local depolyment

ahoffer:
This PR restores local development functionality to that level it was when it was first released. 

- Change local deployment configs to use the service name associated with a container because that is the name that Docker puts into DNS.
 - Change local deployment configs to use the port number exposed in the Dockerfile for communication on the same docker overlay network.
- Remove the `hostname:` attribute from docker compose files because it is misleading. It only changes the hostname internally -- what is returned when executing `hostname` from the command line.

Testing:
Do: Build and deploy locally. Try the ingest, store, and retrieve endpoints.
Expect:
- Store and retrieve should work.
- Ingest should successfully store a product and get as far as sending a call to transform

I tried these same tests in the cloud to see if removing "hostname:" had any negative consequences. I got the results I expected.  The PR hero might want to double-check that and also test it in the cloud.

### DIFF
--- a/deployments/local-minio/configs/mis_config.yml
+++ b/deployments/local-minio/configs/mis_config.yml
@@ -8,4 +8,4 @@ solr:
   host: solr
   port: 8983
 endpointUrl:
-  retrieve: http://cdr-ingest:9041/mis/product/
+  retrieve: http://multi-int-store:9041/mis/product/
--- a/deployments/local-minio/configs/mis_config.yml
+++ b/deployments/local-minio/configs/mis_config.yml
@@ -8,4 +8,4 @@ solr:
   host: solr
   port: 8983
 endpointUrl:
-  retrieve: http://multi-int-store:9041/mis/product/
+  retrieve: http://multi-int-store:8080/mis/product/
--- a/deployments/local-minio/configs/mis_config.yml
+++ b/deployments/local-minio/configs/mis_config.yml
@@ -8,4 +8,4 @@ solr:
   host: solr
   port: 8983
 endpointUrl:
-  retrieve: http://multi-int-store:8080/mis/product/
+  retrieve: http://localhost:9041/mis/product/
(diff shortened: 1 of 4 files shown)


deployments/local-minio/configs/mis_config.yml
```diff
@@ -8,4 +8,4 @@ solr:
   host: solr
   port: 8983
 endpointUrl:
-  retrieve: http://cdr-ingest:9041/mis/product/
\ No newline at end of file
+  retrieve: http://multi-int-store:8080/mis/product/
```
emmberk: This should be updated to `http://localhost:9041/mis/product/`. This config value is only used to create the URIs that are returned by the query endpoint, and these URIs should be available to hit on a browser.

emmberk: I pushed a commit to address this.

emmberk:
rebased and addressed comments

build now

deployments/local-minio/configs/mis_config.yml
```diff
@@ -8,4 +8,4 @@ solr:
   host: solr
   port: 8983
 endpointUrl:
-  retrieve: http://multi-int-store:8080/mis/product/
\ No newline at end of file
+  retrieve: http://localhost:9041/mis/product/
```
ahoffer: Ah! OK. I misunderstood how we were using that property.

emmberk: Actually, you're right. We should send http://multi-int-store:8080/mis/product/ to the transform endpoint, but the query endpoint should return uris with `http://localhost:9041/mis/product/`. I think we should keep this change until we integrate with a transform service for local deployment because it makes testing easier. At that point, we can address the issue of the local hostnames.

deployments/local-minio/configs/mis_config.yml
emmberk: I was wrong when I said that is the only use. See https://github.com/connexta/multi-int-store/pull/119#discussion_r308919845.

emmberk:
build now

PR #122, opened 2019-07-30
Remove unneeded params from store endpoints

emmberk:


multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
```diff
@@ -44,16 +44,16 @@ public S3StorageAdaptor(
 
   @Override
   public void store(
-      @NotBlank final String mimeType,
-      @NotNull final InputStream inputStream,
       @NotNull @Min(1L) @Max(10737418240L) final Long fileSize,
+      @NotBlank final String contentType,
```
jhunzik: Content Type generally refers to the HTTP header. I think in this context it makes more sense to use Media Type (formerly MIME Type https://www.iana.org/assignments/media-types/media-types.xhtml)

multi-int-store/src/main/java/com/connexta/multiintstore/common/MetadataStorageManager.java
```diff
@@ -29,24 +30,35 @@ public MetadataStorageManager(@NotNull final Dao<IndexedProductMetadata, String>
...
-      @NotBlank String productId,
-      @NotBlank String metadataType,
-      @NotBlank String mimeType,
-      @NotNull InputStream inputStream,
-      @NotBlank String fileName)
-      throws IOException, StorageException {
-    if (metadataType.equals(INDEXED_PRODUCT_METADATA_CALLBACK_TYPE)) {
-      final IndexedProductMetadata indexedProductMetadata =
+      @NotBlank final String productId,
+      @NotBlank final String metadataType,
```
ahoffer: :pencil2: Could use `@Pattern(regex="cst")` to validate metadata type.

PR #127, opened 2019-08-01
 Prevent storing of metadata that does not have an associated product

jhunzik:
I added a check to ensure that we only store Metadata if we have a product that has the same ID in S3. This PR also clears out Solr between itests to ensure that stored metadata doesn't bleed over into the next test.

### DIFF
--- a/multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
+++ b/multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
@@ -120,6 +120,6 @@ public class S3StorageAdaptor implements StorageAdaptor {
 
   @Override
   public boolean objectExists(String key) {
-    return (amazonS3.doesObjectExist(bucket, key)) ? true : false;
+    return amazonS3.doesObjectExist(bucket, key);
   }
 }
--- a/multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
+++ b/multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
@@ -14,16 +14,15 @@ import com.amazonaws.services.s3.model.S3Object;
 import com.amazonaws.services.s3.transfer.TransferManager;
 import com.amazonaws.services.s3.transfer.Upload;
 import com.connexta.multiintstore.common.exceptions.StorageException;
-import lombok.extern.slf4j.Slf4j;
-import org.springframework.http.MediaType;
-import org.springframework.util.StringUtils;
-
+import java.io.IOException;
+import java.io.InputStream;
 import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
-import java.io.IOException;
-import java.io.InputStream;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.MediaType;
+import org.springframework.util.StringUtils;
 
 @Slf4j
 public class S3StorageAdaptor implements StorageAdaptor {
@@ -67,7 +66,7 @@ public class S3StorageAdaptor implements StorageAdaptor {
     } catch (RuntimeException | InterruptedException e) {
       throw new StorageException(
           String.format(
-              "Unable to store \"%s\" in bucket \"%s\" with key \"%s\"", fileName, key, bucket),
+              "Unable to store \"%s\" in bucket \"%s\" with key \"%s\"", fileName, bucket, key),
           e);
     }
 
--- a/multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
+++ b/multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
@@ -14,15 +14,16 @@ import com.amazonaws.services.s3.model.S3Object;
 import com.amazonaws.services.s3.transfer.TransferManager;
 import com.amazonaws.services.s3.transfer.Upload;
 import com.connexta.multiintstore.common.exceptions.StorageException;
-import java.io.IOException;
-import java.io.InputStream;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.MediaType;
+import org.springframework.util.StringUtils;
+
 import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
-import lombok.extern.slf4j.Slf4j;
-import org.springframework.http.MediaType;
-import org.springframework.util.StringUtils;
+import java.io.IOException;
+import java.io.InputStream;
 
 @Slf4j
 public class S3StorageAdaptor implements StorageAdaptor {
@@ -116,5 +117,10 @@ public class S3StorageAdaptor implements StorageAdaptor {
 
       throw t;
     }
+  }
+
+  @Override
+  public boolean objectExists(String key) {
+    return (amazonS3.doesObjectExist(bucket, key)) ? true : false;
   }
 }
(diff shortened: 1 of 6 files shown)


jhunzik:
build now

multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
```diff
@@ -117,4 +117,9 @@ public RetrieveResponse retrieve(@NotBlank final String key) throws StorageExcep
       throw t;
     }
   }
+
+  @Override
+  public boolean objectExists(String key) {
+    return (amazonS3.doesObjectExist(bucket, key)) ? true : false;
```
ahoffer: :exclamation:  `doesObjectExist` returns a boolean. No need for the ternary expression.

multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
```diff
@@ -66,7 +66,7 @@ public void store(
     } catch (RuntimeException | InterruptedException e) {
       throw new StorageException(
           String.format(
-              "Unable to store \"%s\" in bucket \"%s\" with key \"%s\"", fileName, key, bucket),
+              "Unable to store \"%s\" in bucket \"%s\" with key \"%s\"", fileName, bucket, key),
```
ahoffer: Good catch!

multi-int-store/src/test/java/com/connexta/multiintstore/SolrTests.java
```diff
@@ -326,13 +335,63 @@ public String getFilename() {
...
+        IOUtils.toInputStream(metadataContents, metadataEncoding);
+    final long metadataFileSize = (long) metadataInputStream.available();
+    final String metadataFileName = "test_file_name.txt";
+    final String metadataMimeType = "text/plain";
+    final MultiValueMap<String, Object> storeMetadataRequestBody = new LinkedMultiValueMap<>();
+    storeMetadataRequestBody.add("fileSize", String.valueOf(metadataFileSize));
+    storeMetadataRequestBody.add("mimeType", metadataMimeType);
+    storeMetadataRequestBody.add(
+        "file",
+        new InputStreamResource(metadataInputStream) {
```
ahoffer: :bulb: We override the contentLength and getFilename methods of InputStreamResource many times in anonymous classes. 

image

Let's create a subclass like:

```
public class KnownInputStreamResource extends InputStreamResource {

    // Pass values into constructor
    public KnownInputStreamResource(InputStream istream, long length, String filename) 


    //Override contentLength() and getFilename()

}
```

 Use this class in place of InputStreamResource

```suggestion
        new KnownInputStreamResource(metadataInputStream, metadataFileSize, metadataFileName);
```

multi-int-store/src/test/java/com/connexta/multiintstore/SolrTests.java
```diff
@@ -326,13 +335,63 @@ public String getFilename() {
...
+            return metadataFileSize;
+          }
+
+          @Override
+          public String getFilename() {
+            return metadataFileName;
+          }
+        });
+    storeMetadataRequestBody.add("fileName", metadataFileName);
+    final HttpHeaders storeMetadataRequestHttpHeaders = new HttpHeaders();
```
ahoffer: :bulb: It is probably time for our own HtttpHeaders factory. This might be something for a different PR, at your discretion.

I imagine it like:

```
@Bean
public class HttpHeadersFactory {

@Value("${acceptVersion}")
private final accept-version;

public static final String ACCEPT_VERSION = "Accept-Version"

public HttpHeaders createWithAcceptVersion() {
    headers = new HttpHeaders();
    headers.set(ACCEPT_VERSION,  accept-version);
    return headers;
}

public static HttpHeaders create() {
    return new HttpHeaders().createWithAcceptVersion();
}

```

Then line  381 becomes
```
  new HttpEntity<>(storeMetadataRequestBody, HttpHeadersFactory.create()));
```

We might even want an HttpEntityFactory that uses the HttpHeadersFactory. The client code would not know that an HttpHeadersFactory existed because it is encapsulated by a HttpEntityFactory

multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
jhunzik: Not sure why I didn't see this when I wrote it. 😂

multi-int-store/src/test/java/com/connexta/multiintstore/SolrTests.java
jhunzik: Good suggestion, I think we should make this a separate PR.

multi-int-store/src/test/java/com/connexta/multiintstore/SolrTests.java
jhunzik: I definitely think it should be a separate PR. I like the idea, but I'm not sure it's worth doing. Can you foresee a time when we have more than the custom `Accept-Version` header.

jhunzik:
build now

jhunzik:
build now

multi-int-store/src/test/java/com/connexta/multiintstore/SolrTests.java
```diff
@@ -93,16 +95,21 @@ public SolrClientConfiguration testSolrClientConfiguration() {
   private String endpointUrlRetrieve;
 
   @Before
-  public void before() {
+  public void before() throws IOException, SolrServerException {
     when(mockAmazonS3.putObject(any(PutObjectRequest.class)))
         .thenReturn(mock(PutObjectResult.class));
+
+    solrClient.deleteByQuery(SOLR_COLLECTION, "*");
+    solrClient.commit(SOLR_COLLECTION);
```
emmberk: We should not need to clear solr between itests. I believe that it is better to write tests that do not have the assumption that Solr is empty, because in a production instance, Solr will never be empty. In each test, we can get the id from the original store request or we can use a fake id to test an id that doesn't exist.

multi-int-store/src/test/java/com/connexta/multiintstore/SolrTests.java
```diff
@@ -326,13 +335,63 @@ public String getFilename() {
   }
 
   @Test
-  @Ignore("TODO verify id matches something in S3 before storing to solr")
-  public void testStoreMetadataProductIdNotFound() {
-    // TODO verify 404? 400?
+  public void testStoreMetadataProductIdNotFound() throws Exception {
+    when(mockAmazonS3.doesObjectExist(anyString(), anyString())).thenReturn(Boolean.TRUE);
```
emmberk: We should be able to stub this method only when the id is the specific one that we expect, not any id.

emmberk: Why do you stub this method twice?

multi-int-store/src/main/java/com/connexta/multiintstore/adaptors/S3StorageAdaptor.java
```diff
@@ -117,4 +117,9 @@ public RetrieveResponse retrieve(@NotBlank final String key) throws StorageExcep
       throw t;
     }
   }
+
+  @Override
+  public boolean objectExists(String key) {
+    return amazonS3.doesObjectExist(bucket, key);
```
emmberk: Why create this new method instead of throwing an exception in `#store` if the id already exists? We should never support overriding, so `#store` should fail for duplicate id regardless of whether `#objectExists` was checked first.

multi-int-store/src/test/java/com/connexta/multiintstore/SolrTests.java
jhunzik: That was a mistake. The stub at the beginning of the method isn't necessary. I'll put up a PR to address your comments.

PR #129, opened 2019-08-01
Convenience changes to local deploy script

ahoffer:
I've been working with local deployments a lot this week.
I made some updates to the deployment script that I find make it more convenient.
I have a better sense of what has to be recreated to pick up different kinds of changes, and how long it takes docker to create (or destroy) things like networks and containers.


ahoffer:
Wrong base branch!

PR #132, opened 2019-08-01
Create the TransferManager in the StorageAdaptor bean

jhunzik:


jhunzik:
build now

jhunzik:
build now

jhunzik:
new day, new build now

ahoffer:
Heroed locally. I cannot hero in the cloud because I am on site. 
I was able to successfully store and retrieve an item.
:heavy_check_mark:

multi-int-store/src/main/java/com/connexta/multiintstore/config/S3StorageConfiguration.java
```diff
@@ -45,18 +43,12 @@ public AmazonS3 amazonS3(@NotNull final AmazonS3Configuration configuration) {
...
-  public TransferManager transferManager(@NotNull AmazonS3 amazonS3) {
-    return TransferManagerBuilder.standard().withS3Client(amazonS3).build();
-  }
-
   @Bean
   public S3StorageAdaptor s3StorageAdaptor(
       @NotNull final AmazonS3 amazonS3,
-      @NotNull TransferManager transferManager,
       @Value("${aws.s3.bucket.quarantine}") @NotBlank final String bucket) {
+    TransferManager transferManager =
```
emmberk: Shouldn't this be in the `S3StorageAdaptor` constructor instead? It's an implementation detail of the `S3StorageAdaptor`, not a configuration.

jhunzik: I think it's better to have Spring provide the object through Dependency Injection rather than the `S3StorageAdaptor` having a dependency directly on `TransferManager`.

emmberk: Only the implementation of S3StorageAdaptor needs TransferManager. If we decide not to use TransferManager in the future to handle storing large files, we should only need to change the S3StorageAdaptor class, not the constructor/configuration.

PR #134, opened 2019-08-05
Allow overwriting of product CST

jhunzik:


multi-int-store/src/main/java/com/connexta/multiintstore/services/impl/IndexedMetadataDao.java
```diff
@@ -31,14 +30,8 @@ public IndexedMetadataDao(IndexedMetadataRepository repository) {
   }
 
   @Override
-  public void save(IndexedProductMetadata indexedProductMetadata)
-      throws StorageException, DuplicateIdException {
+  public void save(IndexedProductMetadata indexedProductMetadata) throws StorageException {
     Optional<IndexedProductMetadata> document = getById(indexedProductMetadata.getId());
```
Kjames5269: ❓ document is no longer used, should it be deleted?

jhunzik: This is why I tagged you 👍

multi-int-store/src/main/java/com/connexta/multiintstore/services/impl/IndexedMetadataDao.java
```diff
@@ -31,14 +30,7 @@ public IndexedMetadataDao(IndexedMetadataRepository repository) {
   }
 
   @Override
-  public void save(IndexedProductMetadata indexedProductMetadata)
```
Kjames5269: ❓ Since all these commands mirror an IndexedMetadataRepository, should we just pass that around instead?

jhunzik:
Closing because we don't want to support overrides and throwing the exception is the correct behavior.

PR #136, opened 2019-08-07
Add Dependabot config file

jhunzik:
We define our dependencies in a unique way, so this may not work right off the bat.


jhunzik:
Closing because I just learned I can configure this in GitHub without having to check in another file.

PR #148, opened 2019-08-09
Default error attributes

ahoffer:
- Rework error handling to use subclasses of Spring's DefaultErrorAttributes class. Instead of writing customer exception handlers, most of the work is done by Spring's own exception handling mechanisms. 
- I did include one global custom error handler class, mostly just to show what it looks like (IngestExceptionHandler).
- Changed some response codes based on conversations with @paouelle 
- Left some notes and TODOs in the code. 
- There are no javadoc comments.
- I didn't put any thought into package names. Also, I think we'd want a place for Patrick's common code outside of this repo.

Thanks to Patrick for the Spring related exception handling classes.


ingest/src/main/java/com/connexta/ingest/client/StoreClient.java
```diff
@@ -71,8 +70,8 @@ public String getFilename() {
     final URI location;
     try {
       location = restTemplate.postForLocation(storeEndpoint, request);
-    } catch (final RestClientException e) {
-      throw new StoreException("Unable to POST to store endpoint", e);
+    } catch (Exception e) {
```
emmberk: Why do we catch all exceptions here? Is it possible that we could get an exception not caused by the Store Service being unavailable? Same question for the TransformClient.

ingest/src/main/java/com/connexta/ingest/controllers/IngestController.java
```diff
@@ -37,19 +37,31 @@ public IngestController(IngestService ingestService) {
...
+          new CouldNotReadAttachment("Ingest unable to read an attachment", e), detail);
+    } catch (IngestException e) {
+      handleIngestException(e, detail);
+    } catch (Exception e) {
+      handleIngestException(new IngestException(detail, e), detail);
     }
     return ResponseEntity.accepted().build();
   }
+
+  private String getDescription(
```
emmberk: ✏️ inline (or make static)

ingest/src/main/java/com/connexta/ingest/controllers/IngestController.java
```diff
@@ -37,19 +37,31 @@ public IngestController(IngestService ingestService) {
...
-          fileName,
-          e);
-      return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build();
+    } catch (IOException e) {
+      handleIngestException(
+          new CouldNotReadAttachment("Ingest unable to read an attachment", e), detail);
+    } catch (IngestException e) {
+      handleIngestException(e, detail);
+    } catch (Exception e) {
+      handleIngestException(new IngestException(detail, e), detail);
```
emmberk: ❓ This is confusing to me. Reminder to ask Aaron about this.

ingest/src/main/java/com/connexta/ingest/exceptions/CouldNotReadAttachment.java
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.ingest.exceptions;
+
+import org.springframework.http.HttpStatus;
+
+public class CouldNotReadAttachment extends IngestException {
```
emmberk: ❓ Why do we need a specific error for this? Isn't this just a bad request?

ingest/src/main/java/com/connexta/ingest/exceptions/StoreServiceNotAvailable.java
```diff
@@ -0,0 +1,16 @@
...
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.ingest.exceptions;
+
+import org.springframework.http.HttpStatus;
+
+public class StoreServiceNotAvailable extends IngestException {
+
+  public StoreServiceNotAvailable(String reason, Throwable cause) {
+    super(HttpStatus.BAD_GATEWAY, reason, cause);
```
emmberk: ❓ Why do we return a 502 here?

ingest/src/main/java/com/connexta/ingest/exceptions/StoreServiceNotAvailable.java
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.ingest.exceptions;
+
+import org.springframework.http.HttpStatus;
+
+public class StoreServiceNotAvailable extends IngestException {
```
emmberk: ❓ Do we care about handling store and tranform client errors differently?

ingest/src/test/java/com/connexta/ingest/IngestApplicationIntegrationTest.java
```diff
@@ -67,6 +68,7 @@
 
   private MockRestServiceServer storeServer;
   private MockRestServiceServer transformServer;
+  private ResultMatcher upstreamServiceUnavailable = status().isBadGateway();;
```
emmberk: ✏️ static

ingest/src/test/java/com/connexta/ingest/controllers/IngestControllerTest.java
```diff
@@ -0,0 +1,30 @@
...
+import static org.mockito.Mockito.mock;
+
+import com.connexta.ingest.exceptions.CouldNotReadAttachment;
+import com.connexta.ingest.service.api.IngestService;
+import java.io.IOException;
+import org.junit.Test;
+import org.mockito.Mockito;
+import org.springframework.web.multipart.MultipartFile;
+
+public class IngestControllerTest {
```
emmberk: ❓ Should we differentiate this class as a unit test class?

ingest/src/main/java/com/connexta/ingest/controllers/IngestController.java
```diff
@@ -37,19 +37,31 @@ public IngestController(IngestService ingestService) {
     String mediaType = multipartFile.getContentType();
```
emmberk: Does this method still need to return a `ResponseEntity`?

deployments/local-minio/configs/ingest_config.yml
```diff
@@ -1,2 +1,19 @@
...
+    org.springframework.web.filter.CommonsRequestLoggingFilter: DEBUG
+    org.springframework.web.client: DEBUG
+management:
+  endpoints:
+    web:
+      exposure:
+        include: "*"
+endpoints:
+  transform:
+    version: "0.1.0" #TODO Use variable substitution from gradle.build file for ${transformApiVersion}
```
emmberk: ❓Does this really need to be added here? I would have expected this to not work before because the local deployment inherits the main docker-compose.yml.

ingest/src/main/java/com/connexta/ingest/client/StoreClient.java
```diff
@@ -71,8 +70,8 @@ public String getFilename() {
...
+      throw new StoreException("Store exception", e);
```
emmberk: ✏️ Move this string to the constructor of StoreException (same for TransformException).
```java
public StoreException(Throwable cause) {
  super("Unable to store", cause);
}
```
```java
public TransformException(Throwable cause) {
  super("Unable to transform", cause);
}
```

ingest/src/main/java/com/connexta/ingest/controllers/IngestController.java
```diff
@@ -37,19 +36,23 @@ public IngestController(IngestService ingestService) {
...
+      inputStream = multipartFile.getInputStream();
+    } catch (IOException e) {
+      throw new ValidationException("Could not open attachment");
     }
+    ingestService.ingest(fileSize, mediaType, inputStream, fileName);
+
     return ResponseEntity.accepted().build();
   }
+
+  private String getDescription(
```
emmberk: ✏️ inline

deployments/local-minio/configs/ingest_config.yml
ahoffer: I'll double check.

emmberk:
build now

emmberk:
`./gradlew clean build` and `./gradlew dependencyCheckAnalyze --info` succeeded for me locally. The CI build failed for a know issue, so I am overriding that build check.

ingest/src/main/java/com/connexta/ingest/client/StoreClient.java
```diff
@@ -71,8 +70,8 @@ public String getFilename() {
     final URI location;
     try {
       location = restTemplate.postForLocation(storeEndpoint, request);
-    } catch (final RestClientException e) {
-      throw new StoreException("Unable to POST to store endpoint", e);
+    } catch (Exception e) {
+      throw new StoreException(e);
```
paouelle: ❓ Why remove the additional info that was added which brings a bit more detail about the condition in which the error occurred?

ingest/src/main/java/com/connexta/ingest/client/TransformClient.java
```diff
@@ -53,8 +53,8 @@ public void requestTransform(
 
     try {
       restTemplate.postForEntity(transformEndpoint, transformRequestHttpEntity, Void.class);
-    } catch (RuntimeException e) {
-      throw new TransformException("Unable to complete transform request", e);
+    } catch (Exception e) {
+      throw new TransformException(e);
```
paouelle: ❓ Same question

ingest/src/main/java/com/connexta/ingest/controllers/IngestController.java
```diff
@@ -37,19 +36,20 @@ public IngestController(IngestService ingestService) {
...
-          "Unable to complete ingest request with params acceptVersion={}, fileSize={}, mediaType={}, fileName={}",
-          acceptVersion,
-          fileSize,
-          mediaType,
-          fileName,
-          e);
-      return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).build();
+      inputStream = multipartFile.getInputStream();
+    } catch (IOException e) {
+      throw new ValidationException("Could not open attachment");
```
paouelle: ❓ Why not piggy back the cause exception? This is a great helper when debugging.

ingest/src/main/java/com/connexta/ingest/exceptions/StoreException.java
```diff
@@ -6,9 +6,9 @@
  */
 package com.connexta.ingest.exceptions;
 
-public class StoreException extends Exception {
+public class StoreException extends IngestException {
 
-  public StoreException(String message, Throwable cause) {
-    super(message, cause);
+  public StoreException(Throwable cause) {
+    super("Store exception", cause);
```
paouelle: ✏️ This seems weird to me as this message is not adding any value beyond the class itself. It is standard to let the client pass its message which would be more context specific or at least some info that is used to be a standard message that is also context specific.

ingest/src/main/java/com/connexta/ingest/exceptions/TransformException.java
```diff
@@ -6,9 +6,9 @@
  */
 package com.connexta.ingest.exceptions;
 
-public class TransformException extends Exception {
+public class TransformException extends IngestException {
 
-  public TransformException(String message, Throwable cause) {
-    super(message, cause);
+  public TransformException(Throwable cause) {
+    super("Transform service exception", cause);
```
paouelle: ✏️ Same comment as in `StoreException`

ingest/src/main/java/com/connexta/ingest/client/StoreClient.java
emmberk: I can provide a a blanket answer to all of your comments in this review:

I suggested to the team in the mob review that we keep things simple and easier to maintain. We were't using a lot of this functionality yet, so I opted to remove it for now and add it back in once we are using these exceptions as intended. We will create tickets to improve the message and details for different scenarios. This PR just adds the groundwork.

paouelle: I am just concerned that this is adding technical debt on some trivial things that should normally be done as you write code. Things like piggy backing a cause exception is part of the guidelines and I am not sure we should skip this as we will definitely forget about that.

PR #160, opened 2019-08-20
Add CircleCI

emmberk:
@connexta/i2o-cicd

### DIFF
--- a/src/test/java/com/connexta/store/CreateProductTests.java
+++ b/src/test/java/com/connexta/store/CreateProductTests.java
@@ -39,7 +39,7 @@ import org.testcontainers.shaded.org.apache.commons.lang.StringUtils;
 @RunWith(SpringRunner.class)
 @SpringBootTest
 @AutoConfigureMockMvc
-public class CreateProductITests {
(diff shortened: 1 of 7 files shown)


emmberk:
build now

emmberk:
# Do not merge
This is blocked by @shaundmorris.

src/test/java/com/connexta/store/CreateProductTests.java
```diff
@@ -39,7 +39,7 @@
 @RunWith(SpringRunner.class)
 @SpringBootTest
 @AutoConfigureMockMvc
-public class CreateProductITests {
+public class CreateProductTests {
```
jhunzik: ❓ Is this the new naming scheme for ITests because they will eventually be in their own Itest directory?

emmberk: No, that was not the reason. I needed a way to only run the tests that use `GenericContainer` on Jenkins and not on CI. An easy way to do this was to rename the tests so that it's easy to skip the *ITest classes. Feel free to update this if you have a better idea.

PR #165, opened 2019-08-22
CircleCI config updates to run itests successfully

aaronilovici:
- Removed docker image for the "Full build without itests" job and used the machine executorType, with an extra step or two to install Java 11
- Changed the name of the job to FullBuild, since it isn't without itests anymore
- Created a single workflow named "default" where both steps run in parallel (dependency can be added with 

The workflow run for this branch (should be at the bottom of the PR anyway, right?): https://circleci.com/workflow-run/6f6a19d1-5320-4375-9ae9-2e9ba773455c


.circleci/config.yml
```diff
@@ -1,17 +1,25 @@
 version: 2.1
 jobs:
-  Full build without itests:
-    docker: &image
-      - image: circleci/openjdk:11-jdk
+  FullBuild:
```
emmberk: ✏️ `Full Build`

.circleci/config.yml
```diff
@@ -1,17 +1,25 @@
...
+            sudo add-apt-repository ppa:openjdk-r/ppa
+            sudo apt-get update -q
+            sudo apt install -y openjdk-11-jdk
+      - run:
+          name: Set default java 11
+          command:
+            sudo update-java-alternatives --set java-1.11.0-openjdk-amd64
       - run:
           name: Building
-          # Skip itests, which require another docker image and don't work in CircleCI
```
emmberk: ❗️The Jenkinsfile should be removed because we no longer need it to run the itests

aaronilovici: All set - removed file

.circleci/config.yml
aaronilovici: All set - renamed

PR #170, opened 2019-09-03
CI/CD Gradle Template

ahoffer:
To test:
1. Look for github checks for circle ci and sonar and that PR cannot be merged without them.
2. Run CI workflow and ensure passes all jobs successfully
3. Check GitHub to ensure checks and badges are updated


emmberk:
We decided to close this PR for now, but we plan to switch to the template soon.

PR #173, opened 2019-09-05
Make the Circle config file conform to the template

jhunzik:


src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
```diff
@@ -45,7 +45,9 @@ public StoreServiceImpl(
       @NotBlank String mediaType,
       @NotBlank String fileName,
       @NotNull InputStream inputStream)
-      throws StoreException, URISyntaxException {
+      throws StoreException, URISyntaxException, InterruptedException {
+    Thread sleepyGuy = new Thread();
+    sleepyGuy.sleep(1000);
```
ahoffer: :question: Was this supposed to be in the commit?

jhunzik: ❗️ Don't commit

PR #189, opened 2019-09-30
Allow compiling with java11+

peterhuffer:
Remove the java version check. Compiling to java 11 should be supported by java 11,12,13,14 based on https://openjdk.java.net/jeps/182

This check is already also done by java and will print out a message like `Could not target platform: 'Java SE 11' using tool chain: 'JDK 8 (1.8)`. if the installed JDK cannot target the target version.

@emmberk @ahoffer @jhunzik


peterhuffer:
build now

PR #203, opened 2019-10-08
Extract version properties

emmberk:


build.gradle
```diff
@@ -33,24 +36,24 @@ repositories {
 }
 
 dependencies {
-    annotationProcessor "org.projectlombok:lombok:1.18.10"
-    compileOnly "org.projectlombok:lombok:1.18.8"
+    annotationProcessor "org.projectlombok:lombok:${lombokVersion}"
+    compileOnly "org.projectlombok:lombok:${lombokVersion}"
     implementation "com.amazonaws:aws-java-sdk-s3:1.11.613"
```
jhunzik: Should we just extract all of these versions as properties for consistency?

emmberk: I prefer to be explicit about versions that are only used in one place.

PR #204, opened 2019-10-08
Extract indexApiVersion

emmberk:


src/main/java/com/connexta/store/config/IndexClientConfiguration.java
```diff
@@ -27,9 +27,10 @@ public RestTemplate nonBufferingRestTemplate() {
   }
 
   @Bean
-  public IndexClient searchClient(
+  public IndexClient indexClient(
       @NotNull @Named("nonBufferingRestTemplate") final RestTemplate restTemplate,
-      @NotBlank @Value("${endpointUrl.index}") final String indexEndpoint) {
-    return new IndexClient(restTemplate, indexEndpoint);
+      @NotBlank @Value("${endpointUrl.index}") final String indexEndpoint,
```
figliold: ❓ Can indexApiVersion be final too?

src/main/java/com/connexta/store/clients/IndexClient.java
```diff
@@ -54,11 +51,10 @@ public String getFilename() {
           }
         });
     final HttpHeaders httpHeaders = new HttpHeaders();
-    // TODO inject IndexApi version
-    httpHeaders.set("Accept-Version", "0.1.0");
+    httpHeaders.set("Accept-Version", indexApiVersion);
```
figliold: ❓ Do you think we should make Accept-Version a constant?

src/test/resources/application.yml
```diff
@@ -3,5 +3,8 @@ server:
 endpointUrl:
   retrieve: http://localhost:${server.port}/mis/product/
   index: http://search:8080/index/
+endpoints:
+  index:
+    version: testIndexApiVersion
```
jhunzik: ❓ Is this supposed to be a placeholder `${testIndexApiVersion}` or an actual value for the version?

emmberk: This is the actual value.

emmberk:
I heroed this locally

PR #205, opened 2019-10-08
Extract storeApiVersion

emmberk:


src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -60,7 +61,15 @@ public StoreController(@NotNull final StoreService storeService) {
...
   public ResponseEntity<Void> storeProduct(
       final String acceptVersion, @Valid final MultipartFile file) {
-    // TODO validate Accept-Version
+    final String expectedAcceptVersion = storeApiVersion;
+    if (!StringUtils.equals(acceptVersion, expectedAcceptVersion)) {
+      log.warn(
+          "Excepted Accept-Version to be \"{}\" but was \"{}\". Only \"{}\" is currently supported.",
+          expectedAcceptVersion,
+          acceptVersion,
+          expectedAcceptVersion);
```
figliold: ❓ Would HTTP 412 Precondition Failed be a better status code here?

emmberk: We should support backwards compatibility, but we currently don't do that. Because of this, I don't think that 412 is appropriate, but maybe there is something better than 501.

ahoffer: In these early days I don't think it matters too much as long as it isn't in the 200s.

emmberk:
I heroed this locally

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -62,7 +63,15 @@ public StoreController(@NotNull final StoreService storeService) {
   @Override
   public ResponseEntity<Void> storeProduct(
       final String acceptVersion, @Valid final MultipartFile file) {
-    // TODO validate Accept-Version
+    final String expectedAcceptVersion = storeApiVersion;
+    if (!StringUtils.equals(acceptVersion, expectedAcceptVersion)) {
+      log.warn(
+          "Excepted Accept-Version to be \"{}\" but was \"{}\". Only \"{}\" is currently supported.",
```
jhunzik: ✏️ `Expected`

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -62,7 +63,15 @@ public StoreController(@NotNull final StoreService storeService) {
   @Override
   public ResponseEntity<Void> storeProduct(
       final String acceptVersion, @Valid final MultipartFile file) {
-    // TODO validate Accept-Version
+    final String expectedAcceptVersion = storeApiVersion;
+    if (!StringUtils.equals(acceptVersion, expectedAcceptVersion)) {
```
peterhuffer: This check is interesting!

peterhuffer: ❓ Just for my understanding, a client will send to this service the expected Store API that it is trying to use?

emmberk: We should support backwards compatibility and non-major version changes, but right now we only support exactly one version. I'm not sure how this should be addressed going forward.

jhunzik: Yeah, I believe it's in the header.

peterhuffer: Would the load balancer determine, based on this version, which service to send it to? Or to spin up a service of that version if one does not exist?

emmberk: 🤷‍♀ I'm not sure how that works, but that sounds right to me.

PR #209, opened 2019-10-17
Returns 404 when retrieving a product that doesn't exist

jhunzik:
This PR also adds a custom error handler similar to the ones we use in the ingest and search services.

Hero:
- Full Build
- Test that we can still successfully retrieve existing products
- Test that we get a 404 when retrieving a product that doesn't exist


src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
```diff
@@ -77,22 +79,23 @@ public void store(
...
     InputStream productInputStream = null;
     try {
-      try {
+      if (amazonS3.doesObjectExist(bucket, key)) {
         s3Object = amazonS3.getObject(new GetObjectRequest(bucket, key));
-      } catch (SdkClientException e) {
-        throw new StoreException("Unable to retrieve product with key " + key, e);
+      } else {
+        throw new RetrieveException(
+            HttpStatus.NOT_FOUND, String.format("Product for key {%s} does not exist", key));
```
emmberk: ✏️ Extend `RetrieveException` instead of specifying the status code here

src/main/java/com/connexta/store/clients/IndexClient.java
```diff
@@ -56,7 +56,8 @@ public String getFilename() {
     try {
       restTemplate.put(indexEndpoint + productId, new HttpEntity<>(body, httpHeaders));
     } catch (Exception e) {
-      throw new StoreException("Error indexing product id=" + productId, e);
+      throw new IndexMetadataException(
+          String.format("Errors indexing product with id {%s}", productId), e);
```
emmberk: ✏️ `error`

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -86,24 +85,23 @@
...
       location = storeService.createProduct(fileSize, mediaType, fileName, inputStream);
     } catch (StoreException | URISyntaxException e) {
-      log.warn(
-          "Unable to complete storeProduct request for a file with mediaType={} and fileName={}",
-          mediaType,
-          fileName,
+      throw new StoreException(
+          String.format(
+              "Unable to complete storeProduct request for a file with mediaType=%s and fileName=%s",
+              mediaType, fileName),
```
emmberk: ✏️ Don't catch StoreException and wrap it in a new StoreException

jhunzik: 🤦‍♂ good catch

src/test/java/com/connexta/store/StoreITests.java
```diff
@@ -277,10 +277,9 @@ public String getFilename() {
     uriBuilder.setPath(uriBuilder.getPath() + "/" + PRODUCT_ID);
 
     // verify
-    // TODO return 404 if key doesn't exist
     assertThat(
         restTemplate.getForEntity(uriBuilder.build(), Resource.class).getStatusCode(),
-        is(HttpStatus.INTERNAL_SERVER_ERROR));
+        is(HttpStatus.NOT_FOUND));
```
emmberk: ✏️ shortcut matcher

PR #212, opened 2019-10-28
Add tests for addMetadata endpoint

emmberk:


src/test/java/com/connexta/store/AddMetadataTests.java
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  @Test
+  public void testAddMetadata() throws Exception {
+    final String datasetId = "341d6c1ce5e0403a99fe86edaed66eea";
+
+    indexMockRestServiceServer
+        .expect(requestTo(endpointUrlIndex + datasetId))
+        .andExpect(method(HttpMethod.PUT))
+        .andRespond(withSuccess());
```
emmberk: ❗️ Verify file is sent with expected attributes

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -106,14 +104,21 @@
...
-  /** TODO Add tests for this endpoint. */
   @Override
   public ResponseEntity<Void> addMetadata(
       final String acceptVersion,
       @Pattern(regexp = "^[0-9a-zA-Z]+$") @Size(min = 32, max = 32) final String productId,
       @Pattern(regexp = "^[0-9a-zA-Z\\-]+$") @Size(min = 1, max = 32) final String metadataType,
       @Valid final MultipartFile file) {
-    // TODO validate params
+    final String expectedAcceptVersion = storeApiVersion;
+    if (!StringUtils.equals(acceptVersion, expectedAcceptVersion)) {
```
peterhuffer: Our API says we should accept previous minor versions, does this need to be updated to account for that?

emmberk: We don't currently support that even though we should. See the message in the exception.

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -64,12 +65,10 @@
...
     if (!StringUtils.equals(acceptVersion, expectedAcceptVersion)) {
-      log.warn(
-          "Expected Accept-Version to be \"{}\" but was \"{}\". Only \"{}\" is currently supported.",
-          expectedAcceptVersion,
-          acceptVersion,
-          expectedAcceptVersion);
-      return ResponseEntity.status(HttpStatus.NOT_IMPLEMENTED).build();
+      throw new UnsupportedOperationException(
+          String.format(
+              "Expected Accept-Version to be \"%s\" but was \"%s\". Only \"%s\" is currently supported.",
```
mcalcote: nano-nit -> Giving the `expectedAcceptVersion` twice seems excessive, `Expected Accept-Version was \"%s\", but only \"%s\" is currently supported.` (and below)

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -129,7 +135,10 @@
       inputStream = file.getInputStream();
     } catch (IOException e) {
       throw new IndexMetadataException(
-          String.format("Unable to read file for PUT CST request for id=%s", productId), e);
+          String.format(
+              "Unable to read file for addMetadata request for metadataType=$s, id=%s",
```
mcalcote: `metadataType=%s, id=%s`?

PR #213, opened 2019-10-28
Index IRM

emmberk:
Adding support for storing and retrieving IRM will be handled in a future PR.

This PR depends on https://github.com/connexta/ion-store/pull/212. The changes for that PR are the first commit.


src/main/java/com/connexta/store/clients/IndexClient.java
```diff
@@ -33,15 +33,15 @@
   @NotBlank private final String indexApiVersion;
 
   public void index(
-      @NotNull final InputStream cstInputStream,
+      @NotNull final InputStream irmInputStream,
```
peterhuffer: ❓ Why change from CST to IRM? Don't we index on CST?

emmberk: We are going to start indexing on IRM instead of CST.

peterhuffer: Temporary or for keeps?

emmberk: Are you suggesting that this variable name be more generic?

peterhuffer: Oh, we are indexing with IRM, but we index only on CST terms right?

emmberk: Correct

peterhuffer: I think it would be fine as a generic `inputStream`

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -64,12 +65,10 @@
...
     if (!StringUtils.equals(acceptVersion, expectedAcceptVersion)) {
-      log.warn(
-          "Expected Accept-Version to be \"{}\" but was \"{}\". Only \"{}\" is currently supported.",
-          expectedAcceptVersion,
-          acceptVersion,
-          expectedAcceptVersion);
-      return ResponseEntity.status(HttpStatus.NOT_IMPLEMENTED).build();
+      throw new UnsupportedOperationException(
+          String.format(
+              "Expected Accept-Version to be \"%s\" but was \"%s\". Only \"%s\" is currently supported.",
```
mcalcote: Could use the constant in the string as well in place of `Accept-Version`. Looks odd to have expectedAcceptVersion twice in the message, but that is just a nit.

src/main/java/com/connexta/store/clients/IndexClient.java
emmberk: I prefer keeping it hardcoded. I think we'll end up having a dynamic set of `data`s in the future where each `data` has a type. As we iterate and support more use cases, I think we can update the type of this parameter, as opposed to the name of this parameter.

src/main/java/com/connexta/store/controllers/StoreController.java
emmberk: I address that in https://github.com/connexta/ion-store/pull/212. This PR should not be merged until https://github.com/connexta/ion-store/pull/212 is merged.

emmberk:
❗️Do not merge before https://github.com/connexta/ion-store/pull/216.

emmberk:
CI will fail until Index API 0.3.0 is released: https://github.com/connexta/ion-index-api/pull/25

emmberk:
I'm not heroing this. I will be heroed in a future PR.

PR #214, opened 2019-10-28
Poller threat model

ahoffer:
I did this with the online app (no linux version of the desktop app). Can someone load it into the desktop app and make sure it renders correctly? 

image


peterhuffer:
Why is the poller threat model depicted here and in https://github.com/connexta/ion-store/pull/215?

mcalcote:
> Why is the poller threat model depicted here and in #215?

In #215 it is marked as out of scope, so it isn't the focus of those threat models.

PR #215, opened 2019-10-28
Updates the Store threat models

jhunzik:
Ingesting:
image

Retrieval:
image

Promotion and Quarantine:
image


peterhuffer:
For retrieval, do we need to show the store's requests going to S3 and cassandra?

peterhuffer:
I might have missed something, but why do we have this PR showing the poller and https://github.com/connexta/ion-store/pull/214 also showing it?

jhunzik:
@peterhuffer 
1. Yeah I'll add that in.
2. We talked about this in the meeting, but both of these services are part of Ion, so we trust them. The store service needs to put the status URL in the queue for the poller to pick up. The delete request should be handled in the poller threat model.
3. I marked it out of scope in my threat models. We should make sure that we don't duplicate threats between the two.

jhunzik:
The last push was just a rebase. Merging and making a note to follow up and add threats and mitigations.

PR #218, opened 2019-10-30
Input validation tests

ahoffer:
Validate input from HTTP requests

### DIFF
--- a/src/main/java/com/connexta/store/controllers/StoreController.java
+++ b/src/main/java/com/connexta/store/controllers/StoreController.java
@@ -6,6 +6,7 @@
  */
 package com.connexta.store.controllers;
 
+import static org.springframework.http.HttpStatus.BAD_REQUEST;
 import static org.springframework.http.ResponseEntity.ok;
 
 import com.connexta.store.adaptors.RetrieveResponse;
@@ -48,6 +49,8 @@ public class StoreController implements StoreApi {
 
   public static final String ACCEPT_VERSION_HEADER_NAME = "Accept-Version";
   public static final String SUPPORTED_METADATA_TYPE = "irm";
+  private static final long GIGABYTE = 1 << 30;
+  public static final long MAX_FILE_BYTES = 10 * GIGABYTE;
 
   @NotNull private final StoreService storeService;
 
@@ -69,19 +72,30 @@ public class StoreController implements StoreApi {
     }
 
     final Long fileSize = file.getSize();
-    // TODO validate that fileSize is (0 GB, 10 GB]
+    if (fileSize > MAX_FILE_BYTES) {
+      throw new CreateDatasetException(
+          BAD_REQUEST,
+          String.format(
+              "File size is %d bytes. File size cannot be greater than %d bytes",
+              fileSize, MAX_FILE_BYTES));
+    }
 
     final String mediaType = file.getContentType();
-    // TODO verify that mediaType is not blank and is a valid Content Type
+    if (mediaType == null) {
+      throw new CreateDatasetException(BAD_REQUEST, "Media type is missing");
+    }
 
     final String fileName = file.getOriginalFilename();
-    // TODO verify that fileName is not blank
+    if (StringUtils.isEmpty(fileName)) {
+      throw new CreateDatasetException(BAD_REQUEST, "Filename is missing");
+    }
 
     final InputStream inputStream;
     try {
       inputStream = file.getInputStream();
     } catch (IOException e) {
       throw new CreateDatasetException(
+          BAD_REQUEST,
           String.format(
               "Unable to read file for createDataset request with mediaType=%s and fileName=%s",
               mediaType, fileName),
--- a/src/main/java/com/connexta/store/controllers/StoreController.java
+++ b/src/main/java/com/connexta/store/controllers/StoreController.java
@@ -6,7 +6,6 @@
  */
 package com.connexta.store.controllers;
 
-import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.springframework.http.ResponseEntity.ok;
 
 import com.connexta.store.adaptors.RetrieveResponse;
@@ -50,8 +49,6 @@ public class StoreController implements StoreApi {
 
   public static final String ACCEPT_VERSION_HEADER_NAME = "Accept-Version";
   public static final String SUPPORTED_METADATA_TYPE = "irm";
-  private static final long GIGABYTE = 1 << 30;
-  public static final long MAX_FILE_BYTES = 10 * GIGABYTE;
 
   @NotNull private final StoreService storeService;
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -6,6 +6,7 @@
  */
 package com.connexta.store.controllers;
 
+import static org.springframework.http.HttpStatus.*;
```
ahoffer: There is something about IntelliJ converting imports to asterisks. Why doesn't Spotless doesn't fix these * imports?  If is bugs people I can go back and use explicit imports.

emmberk: I think we should use explicit imports.

emmberk: Here are my settings in Intellij
image

figliold: I believe the company coding standards state using explicit imports.

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -48,6 +49,8 @@
 
   public static final String ACCEPT_VERSION_HEADER_NAME = "Accept-Version";
   public static final String SUPPORTED_METADATA_TYPE = "cst";
+  private static final long GIGABYTE = 1 << 30;
+  public static final long MAX_FILE_BYTES = 10 * GIGABYTE;
```
emmberk: ✏️ `private`

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -69,19 +72,30 @@
     }
 
     final Long fileSize = file.getSize();
-    // TODO validate that fileSize is (0 GB, 10 GB]
+    if (fileSize > MAX_FILE_BYTES) {
+      throw new CreateDatasetException(
```
emmberk: ✏️ `throw new ValidationException`

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -69,19 +72,30 @@
...
+          BAD_REQUEST,
+          String.format(
+              "File size is %d bytes. File size cannot be greater than %d bytes",
+              fileSize, MAX_FILE_BYTES));
+    }
 
     final String mediaType = file.getContentType();
-    // TODO verify that mediaType is not blank and is a valid Content Type
+    if (mediaType == null) {
```
emmberk: ✏️  If the mediaType is not null could it still be invalid? If so, I think it should be validated here, or we should keep that TODO.

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -69,19 +72,30 @@
...
+      throw new CreateDatasetException(BAD_REQUEST, "Media type is missing");
```
emmberk: ✏️ `throw new ValidationException`

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -69,19 +72,30 @@
...
+    }
 
     final String fileName = file.getOriginalFilename();
-    // TODO verify that fileName is not blank
+    if (StringUtils.isEmpty(fileName)) {
+      throw new CreateDatasetException(BAD_REQUEST, "Filename is missing");
```
emmberk: ✏️ `throw new ValidationException`

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -69,19 +72,30 @@
...
 
     final String mediaType = file.getContentType();
-    // TODO verify that mediaType is not blank and is a valid Content Type
+    if (mediaType == null) {
+      throw new CreateDatasetException(BAD_REQUEST, "Media type is missing");
+    }
 
     final String fileName = file.getOriginalFilename();
-    // TODO verify that fileName is not blank
+    if (StringUtils.isEmpty(fileName)) {
```
emmberk: ✏️ `StringUtils.isBlank`

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -69,19 +72,30 @@
...
+      throw new CreateDatasetException(BAD_REQUEST, "Filename is missing");
+    }
 
     final InputStream inputStream;
     try {
       inputStream = file.getInputStream();
     } catch (IOException e) {
       throw new CreateDatasetException(
```
emmberk: ✏️ `throw new ValidationException`

src/test/java/com/connexta/store/CreateDatasetTests.java
```diff
@@ -18,28 +25,34 @@
...
+import org.springframework.web.multipart.MultipartFile;
+import org.springframework.web.server.ResponseStatusException;
 
 @SpringBootTest
 @AutoConfigureMockMvc
 public class CreateDatasetTests {
 
   @MockBean private AmazonS3 mockAmazonS3;
 
+  @Inject private StoreController storeController;
```
emmberk: ✏️ Tests in this class are using MockMvc. I think that tests that test StoreController directly belong in a different class.

src/test/java/com/connexta/store/CreateDatasetTests.java
```diff
@@ -89,9 +102,40 @@ public void testBadAcceptVersion() throws Exception {
...
-    // TODO verify 400
+  void testCantReadAttachment() throws IOException {
+    MultipartFile file = spy(newValidMultipartFile());
+    doThrow(new IOException("Cannot read attachment")).when(file).getInputStream();
+    assertBadRequest(file);
+    verify(file).getInputStream();
+  }
+
+  @Test
+  void testFileTooLarge() {
```
peterhuffer: ❓ What happens when a negative file size is passed in?

src/main/java/com/connexta/store/controllers/StoreController.java
ahoffer: ValidateException. That would have been good to know before I got started. Let's put hints, tips, design constraints, or technical expectations into the comments on tickets.

src/main/java/com/connexta/store/controllers/StoreController.java
peterhuffer: ❓ Do we need to do any sort of escaping on these inputs?

ahoffer: Not that I know of.

src/test/java/com/connexta/store/CreateDatasetTests.java
ahoffer: Size comes from the stream and streams cannot have a negative length.

src/main/java/com/connexta/store/controllers/StoreController.java
ahoffer: AFAIK we except any media type. At this stage in the application, the File is just a stream of bits. We store it and ask transform to interpret it. If transform can interpret the File enough to know if it can be promoted, then all is well.

ahoffer: I'll go ahead and add a test for empty or blank strings, but I think they are unlikely. These kinds of things are either null or well-formed. Usually.

PR #219, opened 2019-10-31
Store IRM and send URI to SearchService

emmberk:
This PR is ready for review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,13 +63,15 @@ For more information about spotless checks see
     s3:
       endpointUrl: http://minio:9000
       region: local
-      bucket: ingest-quarantine
+      bucket:
+        file: ingest-quarantine
+        irm: irm
     ```
 
     Example configs/store_config.yml:
     ```yaml
     endpointUrl:
-      retrieve: http://localhost:9041/dataset/
+      retrieve: http://localhost:9041
     ```
 
     The service is capable of storing data in an S3-compatible data store.
--- a/src/main/java/com/connexta/store/adaptors/FileRetrieveResponse.java
+++ b/src/main/java/com/connexta/store/adaptors/FileRetrieveResponse.java
@@ -13,7 +13,7 @@ import lombok.Data;
 import org.springframework.http.MediaType;
 
 @Data
-public class RetrieveResponse {
+public class FileRetrieveResponse {
 
   @NotNull private final MediaType mediaType;
   @NotNull private final InputStream inputStream;
--- a/src/main/java/com/connexta/store/adaptors/StorageAdaptorRetrieveResponse.java
+++ b/src/main/java/com/connexta/store/adaptors/StorageAdaptorRetrieveResponse.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.store.adaptors;
+
+import java.io.InputStream;
+import java.util.Map;
+import javax.validation.constraints.NotNull;
+import lombok.Data;
+import org.springframework.http.MediaType;
+
+@Data
+public class StorageAdaptorRetrieveResponse {
+
+  @NotNull private final MediaType mediaType;
+  @NotNull private final InputStream inputStream;
+  @NotNull private final Map<String, String> metadata;
+}
--- a/src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
+++ b/src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
@@ -6,16 +6,18 @@
  */
 package com.connexta.store.service.impl;
 
-import com.connexta.store.adaptors.RetrieveResponse;
+import com.connexta.store.adaptors.FileRetrieveResponse;
 import com.connexta.store.adaptors.StorageAdaptor;
+import com.connexta.store.adaptors.StorageAdaptorRetrieveResponse;
 import com.connexta.store.clients.IndexDatasetClient;
-import com.connexta.store.exceptions.CreateDatasetException;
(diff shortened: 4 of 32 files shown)


README.md
```diff
@@ -63,7 +63,9 @@ For more information about spotless checks see
     s3:
       endpointUrl: http://minio:9000
       region: local
-      bucket: ingest-quarantine
+      bucket:
```
ahoffer: Someday we may decide that one kind of bucket is fine for any type of Data of we want to persist to S3. Someday.

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
```diff
@@ -25,42 +28,71 @@
...
       @Pattern(regexp = "^[0-9a-zA-Z]+$") @Size(min = 32, max = 32) final String datasetId)
-      throws IndexMetadataException {
+      throws IndexDatasetException, URISyntaxException {
     // TODO check that the dataset exists
-    // TODO store IRM
-    indexDatasetClient.indexDataset(irmInputStream, fileSize, datasetId);
+    irmStorageAdaptor.store(
+        fileSize, StoreController.IRM_MEDIA_TYPE_VALUE, irmInputStream, datasetId, Map.of());
+    log.info("Successfully stored IRM for datasetId={}", datasetId);
+    indexDatasetClient.indexDataset(
```
ahoffer: At some point in time we should rethink where we call `indexDatasetClient`. It is a method on StoreManager, but it isn't really being stored in the the same way a File is stored. The Metadata is being parsed and made searchable. Some of it (maybe all?) will end up persisted in s search engine. 

I think the right place to call `indexDatasetClient` is from the controllers. Controllers are a natural place to to use an HTTP client. Controllers take data and compose it to create domain objects (File, Metacard, ...), then pass those objects into the service layer.

src/main/java/com/connexta/store/adaptors/FileRetrieveResponse.java
```diff
@@ -13,7 +13,7 @@
 import org.springframework.http.MediaType;
 
 @Data
-public class RetrieveResponse {
+public class FileRetrieveResponse {
```
ahoffer: Looking at the fields of FileRetrieveResponse: inputstream, mediatype, and name. This is a model of a File. It is created from a response, but isn't a Response. A good name for the class might be `File`, `IonFile`, or `DataFile`  because that is what it represents.

src/main/java/com/connexta/store/adaptors/StorageAdaptorRetrieveResponse.java
```diff
@@ -0,0 +1,21 @@
...
+package com.connexta.store.adaptors;
+
+import java.io.InputStream;
+import java.util.Map;
+import javax.validation.constraints.NotNull;
+import lombok.Data;
+import org.springframework.http.MediaType;
+
+@Data
+public class StorageAdaptorRetrieveResponse {
```
ahoffer: Similar observation about naming this class.This class models an S3Object. It is very much like a File, but has a richer set of metadata than just filename. A good name might be DataObject, IonOjbect, or Blob (Binary Large OBject).

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
```diff
@@ -25,42 +28,71 @@
...
-  public @NotNull RetrieveResponse retrieveFile(
+  public @NotNull FileRetrieveResponse retrieveFile(
       @Pattern(regexp = "^[0-9a-zA-Z]+$") @Size(min = 32, max = 32) final String datasetId)
       throws RetrieveException {
-    return storageAdaptor.retrieve(datasetId);
+    final StorageAdaptorRetrieveResponse storageAdaptorRetrieveResponse =
+        fileStorageAdaptor.retrieve(datasetId);
+
+    final String fileName =
+        storageAdaptorRetrieveResponse.getMetadata().get(FILE_NAME_METADATA_KEY);
```
ahoffer: Encapsulate the knowledge of how the metadata is organized inside the StorageAdaptorRetrieveResponse class. Don't expose the information about the metadata map to other classes. 

```
StorageAdaptorRetrieveResponse

String getFilename() {
   String filename = metadata.get(FILE_NAME_METADATA_KEY);
       if (StringUtils.isEmpty(filename)) {
      throw new RetrieveException(
          String.format(
              "Expected S3 object to have a non-null metadata value for %s",
              FILE_NAME_METADATA_KEY));
     }
    return filename;
}
```

src/main/java/com/connexta/store/adaptors/StorageAdaptorRetrieveResponse.java
ahoffer: Turn the FileRetrieveResponse class into interface. Have the StorageAdaptorRetrieveResponse implement the interface.

emmberk:
Rebased on top of https://github.com/connexta/ion-store/pull/222. Still need to fix/write tests.

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
emmberk: This comment might be outdated. We have been following the structure of each Controller has a 1 corresponding Service. IndexDatasetClient is only used in the StoreServiceImpl. I think Clients should only deal with the REST/spring stuff and not contain any app logic.

emmberk:
I have heroed this locally.

jhunzik:
I heroed this locally along with the changes in https://github.com/connexta/ion-search/pull/66.
The hero included:
- Happy path through ingest, simulated PUTting IRM to the store, indexing IRM, retrieving IRM, retrieving the file, and querying for IRM result URLs.
- Correct exceptions when any of the requests are invalid.
- Store returns 500 when search or solr is down while trying to index
- Search returns correct error JSON when Solr is down

PR #223, opened 2019-11-06
Polling service

ahoffer:
[READY FOR REVIEW]
This is the proposed structure of the polling service's packages, classes, and interfaces.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,6 @@
+#Wed Nov 06 08:06:17 MST 2019
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.2-all.zip
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.2-bin.zip
-zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
--- a/src/main/java/com/connexta/poller/api/PollingService.java
+++ b/src/main/java/com/connexta/poller/api/PollingService.java
@@ -8,10 +8,6 @@ package com.connexta.poller.api;
 
 import java.util.concurrent.Future;
 
-/**
- * This service accepts a PollingTask and returns a Future. The Future provides the results of the
- * polling task.
- */
 public interface PollingService {
 
   <T> Future<T> poll(PollingTask<T> pollingTask);
--- a/src/main/java/com/connexta/poller/api/PollingService.java
+++ b/src/main/java/com/connexta/poller/api/PollingService.java
@@ -1,0 +1,18 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.poller.api;
+
+import java.util.concurrent.Future;
+
+/**
+ * This service accepts a PollingTask and returns a Future. The Future provides the results of the
+ * polling task.
+ */
+public interface PollingService {
+
+  <T> Future<T> poll(PollingTask<T> pollingTask);
+}
--- a/src/main/java/com/connexta/poller/api/PollingTask.java
+++ b/src/main/java/com/connexta/poller/api/PollingTask.java
@@ -1,0 +1,17 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.poller.api;
+
+import com.dyngr.core.AttemptMaker;
+import com.dyngr.core.AttemptResult;
+
+/** This interface is the boundary between the polling library and client code. */
+public interface PollingTask<T> extends AttemptMaker<T> {
+
+  // AttemptResult is a class from the open source polling library
+  AttemptResult<T> process();
+}
--- a/src/main/java/com/connexta/poller/api/PollingTask.java
+++ b/src/main/java/com/connexta/poller/api/PollingTask.java
@@ -9,7 +9,7 @@ package com.connexta.poller.api;
 import com.dyngr.core.AttemptMaker;
 import com.dyngr.core.AttemptResult;
 
-/** This interface is the boundary between the polling library and client code. */
+/** This interface is a boundary between the polling library and client code. */
 public interface PollingTask<T> extends AttemptMaker<T> {
 
   // AttemptResult is a class from the open source polling library
--- a/src/main/java/com/connexta/poller/service/TransformPollingService.java
+++ b/src/main/java/com/connexta/poller/service/TransformPollingService.java
@@ -8,6 +8,7 @@ package com.connexta.poller.service;
 
 import com.connexta.poller.api.PollingService;
 import com.connexta.poller.api.PollingTask;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Future;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
@@ -24,6 +25,6 @@ public class TransformPollingService implements PollingService {
 
   @Override
   public <T> Future<T> poll(PollingTask<T> pollingTask) {
-    return null;
+    return new CompletableFuture<>();
   }
 }
--- a/src/main/java/com/connexta/poller/service/TransformPollingService.java
+++ b/src/main/java/com/connexta/poller/service/TransformPollingService.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.poller.service;
+
+import com.connexta.poller.api.PollingService;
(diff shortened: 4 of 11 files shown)


ahoffer:
>  How was the dyngr library was selected?

I started by googling for "java polling library". Much of what I found had a "role your own" feel because retries and timeouts were not provided. This library included those features. A bonus was that library does not bring in any additional runtime dependencies.

src/main/java/com/connexta/poller/api/PollingService.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.poller.api;
+
+import java.util.concurrent.Future;
+
+/**
+ * This service accepts a PollingTask and returns a Future. The Future provides the results of the
```
peterhuffer: ✏️ This class javadoc is just describing the method. The javadoc should explain the PollingService's responsibilities in the polling API.

src/main/java/com/connexta/poller/api/PollingTask.java
```diff
@@ -0,0 +1,17 @@
...
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.poller.api;
+
+import com.dyngr.core.AttemptMaker;
+import com.dyngr.core.AttemptResult;
+
+/** This interface is the boundary between the polling library and client code. */
+public interface PollingTask<T> extends AttemptMaker<T> {
```
peterhuffer: ❓ Why extend this interface rather than just used the AttemptMaker directly? These both define the same method. We are already coupling our polling API to the library by returning the AttemptResult below.
I'd argue that we should not be exposing a third party library in our API.

src/main/java/com/connexta/poller/service/TransformPollingService.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * This is class is meant to be instantiated as a Bean, Component, or Service. The class's
+ * responsibilities are to manage the settings for polling and kick off new polling tasks. Other
+ * responsibilities include settings the wait and retry behavior. It also creates and configures the
+ * executor, and ensures that it is shutdown properly.
+ */
+@Service
+@Slf4j
+public class TransformPollingService implements PollingService {
```
peterhuffer: ❓ Do we really need a polling interfaces? I think this class would do just fine as a bean with an internal timer on when it runs. I would imagine that we would not expect any extending of our system's capabilities through these interfaces.

src/main/java/com/connexta/poller/service/TransformPollingTask.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * This polling task implements the logic to contact the transform status endpoint and receive
+ * status information about a specific job. The polling task will complete when the transform job
+ * reaches the complete or failed status. The polling task will also end if the PollingService
+ * exceeds the services maximum number of number of retries or reaches other service-specific
+ * thresholds.
+ *
+ * @param <T>
+ */
+public class TransformPollingTask<T> implements PollingTask<T> {
```
peterhuffer: Same question about the API.

src/test/java/com/connexta/poller/service/TransformPollingServiceUnitTests.java
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.poller.service;
+
+import org.junit.jupiter.api.Test;
+
+class TransformPollingServiceUnitTests {
```
peterhuffer: ✏️ `TransformPollingServiceTest`

src/test/java/com/connexta/poller/service/TransformPollingTaskUnitTests.java
```diff
@@ -0,0 +1,17 @@
...
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.poller.service;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.junit.jupiter.api.Test;
+
+class TransformPollingTaskUnitTests {
```
peterhuffer: ✏️ `TransformPollingTaskTest`

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Wed Nov 06 08:06:17 MST 2019
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.2-all.zip
```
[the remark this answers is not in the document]
ahoffer: It was generated. Should I leave it as-is or roll it back?

src/main/java/com/connexta/poller/api/PollingService.java
ahoffer: I'm struggling to come up with a good class comment.  
_A service that executes a poll task and a returns a value_
What can I said about it that isn't obvious?
I just removed the commend for now.

gradle/wrapper/gradle-wrapper.properties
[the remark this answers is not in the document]
peterhuffer: Leaving it is good as long as it's generated. Just wanted to make sure.

src/main/java/com/connexta/poller/api/PollingTask.java
ahoffer: I went back and forth on it. On the one hand, adding this interface and an impl adds more code and a layer of indirection (added complexity). On the other hand, the PollingTask interface and impl completely insulate other Ion objects from the polling library. AttemptResult can AttemptMaker end up being 100% internal to the PollingTask.  I contemplated which approach was more Java-like. I decided that adding the interface was more Java-like. That said, Java's idioms have never made a lot of sense to me. Things that are best practice in Java are not necessarily best practices in other contexts. If the team feels that another approach is more inline with Java best practices, I''ll gladly change it.

peterhuffer: Which other Ion objects are interacting with this polling? I figured it would just be a thread that kicks off on system startup and does it's thing for the duration that the process is alive.

src/main/java/com/connexta/poller/api/PollingService.java
peterhuffer: Which piece of code will be calling the `poll` method and managing the Futures?

src/main/java/com/connexta/poller/service/TransformPollingService.java
ahoffer: I don't know. I've been dinged in the past for not following Java best practices.  Java best practices includes lots of interfaces. In the long run I don't think using the interfaces or not using the interfaces is going to  make a big difference. Other people may have stronger opinions on the matter and I'm totally open to that. I wish GitHub had surveys for this kind of thing : "(A) The interfaces stay. (B) The interfaces go."

src/main/java/com/connexta/poller/service/TransformPollingTask.java
ahoffer: https://github.com/connexta/ion-store/pull/223/files#r343311328

src/test/java/com/connexta/poller/service/TransformPollingServiceUnitTests.java
ahoffer: I saw a PR yesterday (don't remember which one) that had classes that test classes that ended in "UnitTests". I thought is was something new we started doing. No?

src/main/java/com/connexta/poller/api/PollingTask.java
ahoffer: I expect there will be a single consumer of the polling service-- the one that polls the transform status endpoint. 

Creating interfaces has been drilled into me time so many times that I can no longer use a coffee machine unless it implements the `HotBeverageFactory<T extends HotBeverage>` interface. :coffee:

src/main/java/com/connexta/poller/api/PollingService.java
ahoffer: That code hasn't been written yet. It won't be in this PR because it is tied to a different ticket. Well... I guess that is up for debate. You can read the other ticket -- it's in the sprint --and decide for yourself.

In my mind, I imagine only a single object, probably a @Service, talking to the poller code.

peterhuffer: I thought this ticket was for determining how polls are executed and the other ticket was the actual poll implementation.

ahoffer:
Peter and I talked about some alternate designs. Instead of a Poller Service, there is a Work Queue. The client code drops a Runnable onto the queue. A Service picks up the Runnable and executes it -- The client doesn't interact with the Poller Service at all. The Runnable handles everything-- the polling, the call getting the IRM, storing it, promoting the dataset and sending the index request.

src/main/java/com/connexta/poller/api/PollingService.java
```diff
@@ -0,0 +1,14 @@
...
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.poller.api;
+
+import java.util.concurrent.Future;
+
+public interface PollingService {
+
+  <T> Future<T> poll(PollingTask<T> pollingTask);
```
peterhuffer: ❓ What happens if there is an error during polling? Like a network failure.

ahoffer:
Changing approach to put a queue in between the Store and Poller

PR #225, opened 2019-11-11
WIP: Ingest merge into Store

mcalcote:
Deployment Local or otherwise has not been evaluated ... yet

This pr is being opened to gather feedback on the current approach.


src/main/java/com/connexta/store/config/TransformClientConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.web.client.RestTemplateBuilder;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.web.client.RestTemplate;
+
+@Configuration
+public class TransformClientConfiguration {
+
+  @Bean("transformClientRestTemplate")
```
ahoffer: :question: Why does everyone get their own RestTemplate object?

src/test/java/com/connexta/store/controllers/StoreControllerTest.java
```diff
@@ -32,11 +33,13 @@
...
   @Mock private StoreService mockStoreService;
 
+  @Mock private IngestService mockIngestService;
+
   private StoreApi storeApi;
 
   @BeforeEach
   public void beforeEach() {
-    storeApi = new StoreController(mockStoreService, STORE_API_VERSION);
+    storeApi = new StoreController(mockIngestService, mockStoreService, STORE_API_VERSION);
```
ahoffer: It might be possible to inject the ingest and store services as `@MockBean`, and then `@Inject` to inject the store controller instead of calling its constructor.

src/test/java/com/connexta/store/service/impl/IngestServiceImplUnitTests.java
```diff
@@ -0,0 +1,318 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InOrder;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+public class IngestServiceImplUnitTests {
```
ahoffer: :pencil2: This class is really long.

PR #226, opened 2019-11-11
Implement how polling tasks are executed

ahoffer:
[Ready for Review  2019-11-14T08:00]

NOTE: I had some problems with the polling thread mysteriously dying. I backed-up a few commits to where it was working and moved forward. I may have lost some changes I made in response to your comments. 

Implements how polling tasks are executed and provides a place to send requests to Store Service for promote/quarantine. Look for "TODO Follow-on ticket" for the place holders.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,6 +13,7 @@ buildscript {
         springBootVersion = "2.2.1.RELEASE"
         springfoxSwaggerVersion = "2.9.2"
         storeApiVersion = "0.3.0"
+        transformApiVersion = "0.2.0"
     }
     repositories {
         maven { url "https://plugins.gradle.org/m2/" }
@@ -40,9 +41,11 @@ repositories {
 dependencies {
     annotationProcessor "org.projectlombok:lombok:${lombokVersion}"
     compileOnly "org.projectlombok:lombok:${lombokVersion}"
+    compileOnly "com.connexta.transformation:transformation-api-rest-spring-stubs:${transformApiVersion}"
     implementation "com.amazonaws:aws-java-sdk-s3:1.11.613"
     implementation "com.connexta.search:index-api-rest-spring-stubs:${indexApiVersion}"
     implementation "com.connexta.store:store-api-rest-spring-stubs:${storeApiVersion}"
+    implementation "com.dyngr:polling:1.1.3"
     implementation "io.springfox:springfox-swagger2:${springfoxSwaggerVersion}"
     implementation "io.springfox:springfox-swagger-ui:${springfoxSwaggerVersion}"
     implementation "javax.inject:javax.inject:1"
@@ -106,7 +109,7 @@ dependencyCheck {
 }
 
 processResources {
-    expand([indexApiVersion: "${indexApiVersion}", storeApiVersion: "${storeApiVersion}"])
+    expand([indexApiVersion: "${indexApiVersion}", storeApiVersion: "${storeApiVersion}",  transformApiVersion: "${transformApiVersion}"])
 }
 
 bootJar {
--- a/src/main/java/com/connexta/poller/service/StatusService.java
+++ b/src/main/java/com/connexta/poller/service/StatusService.java
@@ -1,0 +1,14 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.poller.service;
+
+import java.net.URI;
+
+public interface StatusService {
+
+  void submit(URI uri);
+}
--- a/src/main/java/com/connexta/poller/service/StatusServiceImpl.java
+++ b/src/main/java/com/connexta/poller/service/StatusServiceImpl.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.poller.service;
+
+import static com.dyngr.core.AttemptResults.continueFor;
+import static com.dyngr.core.AttemptResults.finishWith;
+import static com.dyngr.core.AttemptResults.justContinue;
+
+import com.dyngr.PollerBuilder;
+import com.dyngr.core.AttemptMaker;
+import com.dyngr.core.StopStrategies;
+import com.dyngr.core.WaitStrategies;
+import java.net.URI;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.web.client.ResourceAccessException;
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * The class's responsibilities are to manage the settings for polling and kick off new polling
+ * tasks. Other responsibilities include settings the wait and retry behavior.
+ */
+@Slf4j
+@AllArgsConstructor
+public class StatusServiceImpl implements StatusService {
+
+  @Min(0)
+  private final int secondsBetweenRetries;
+
+  @Min(1)
+  private final int secondsToLive;
+
+  @NotNull private final ExecutorService executorService;
+  @NotNull private final RestTemplate restTemplate;
+
+  @Override
+  @SuppressWarnings("unused")
+  public void submit(URI uri) {
+    StatusResponse status = null;
+    try {
+      status = poll(uri);
+    } catch (ExecutionException | InterruptedException e) {
+      log.info("Polling failed for {}", uri);
+      return;
+    }
+    process(status);
+  }
+
+  private StatusResponse poll(URI uri) throws ExecutionException, InterruptedException {
+    Future<StatusResponse> statusFuture = builder().polling(getAttemptMaker(uri)).build().start();
+    return statusFuture.get();
+  }
+
+  private AttemptMaker<StatusResponse> getAttemptMaker(URI uri) {
+    return () -> {
+      StatusResponse statusResponse = null;
+      try {
+        statusResponse = restTemplate.getForObject(uri, StatusResponse.class);
+      } catch (ResourceAccessException e) {
+        continueFor(e);
+      }
+
+      // TODO: If the transform job is done, stop polling
+      if (statusResponse != null && "complete".equalsIgnoreCase(statusResponse.getStatus())) {
+
+        // Stop polling
+        return finishWith(statusResponse);
+      }
+
+      // Job not done, keep polling
+      return justContinue();
+    };
+  }
+
+  protected void process(StatusResponse status) {
+    // TODO: Handle transformation
+  }
+
+  private PollerBuilder builder() {
+    return PollerBuilder.newBuilder()
+        .withExecutorService(executorService)
+        .stopIfException(false)
+        .withWaitStrategy(WaitStrategies.fixedWait(secondsBetweenRetries, TimeUnit.SECONDS))
+        .withStopStrategy(StopStrategies.stopAfterDelay(secondsToLive, TimeUnit.SECONDS));
+  }
+}
--- a/src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
+++ b/src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
@@ -6,6 +6,7 @@
  */
 package com.connexta.store.service.impl;
 
+import com.connexta.poller.service.StatusService;
 import com.connexta.store.adaptors.FileRetrieveResponse;
 import com.connexta.store.adaptors.StorageAdaptor;
 import com.connexta.store.adaptors.StorageAdaptorRetrieveResponse;
@@ -40,6 +41,7 @@ public class StoreServiceImpl implements StoreService {
   @NotNull private final StorageAdaptor fileStorageAdaptor;
   @NotNull private final StorageAdaptor irmStorageAdaptor;
   @NotNull private final IndexDatasetClient indexDatasetClient;
+  @NotNull private final StatusService statusService;
 
   @Override
   @NotNull
@@ -52,6 +54,10 @@ public class StoreServiceImpl implements StoreService {
     final String datasetId = UUID.randomUUID().toString().replace("-", "");
     fileStorageAdaptor.store(
         fileSize, mediaType, fileInputStream, datasetId, Map.of(FILE_NAME_METADATA_KEY, fileName));
+
+    // TODO: Call the transform status service
+    // statusService.submit(uri);
+
     return UriComponentsBuilder.fromUri(retrieveUri)
         .path(StoreController.RETRIEVE_FILE_URL_TEMPLATE)
         .build(datasetId);
--- a/src/main/java/com/connexta/poller/service/StatusResponse.java
+++ b/src/main/java/com/connexta/poller/service/StatusResponse.java
@@ -1,0 +1,12 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.poller.service;
+
+public interface StatusResponse {
+
+  String getStatus();
+}
--- a/src/main/java/com/connexta/poller/service/StatusResponseImpl.java
+++ b/src/main/java/com/connexta/poller/service/StatusResponseImpl.java
@@ -1,0 +1,15 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.poller.service;
+
+import lombok.Data;
+
+@Data
+public class StatusResponseImpl implements StatusResponse {
+
+  private final String status;
+}
--- a/src/main/java/com/connexta/store/config/StoreServiceConfiguration.java
+++ b/src/main/java/com/connexta/store/config/StoreServiceConfiguration.java
@@ -6,6 +6,7 @@
  */
 package com.connexta.store.config;
 
+import com.connexta.poller.service.StatusService;
 import com.connexta.store.adaptors.StorageAdaptor;
 import com.connexta.store.clients.IndexDatasetClient;
 import com.connexta.store.service.api.StoreService;
@@ -27,9 +28,14 @@ public class StoreServiceConfiguration {
       @NotBlank @Value("${endpointUrl.retrieve}") final String retrieveEndpoint,
       @NotNull @Named("fileStorageAdaptor") final StorageAdaptor fileStorageAdapter,
       @NotNull @Named("irmStorageAdaptor") final StorageAdaptor irmStorageAdapter,
-      @NotNull final IndexDatasetClient indexDatasetClient)
+      @NotNull final IndexDatasetClient indexDatasetClient,
+      @NotNull final StatusService statusService)
       throws URISyntaxException {
     return new StoreServiceImpl(
-        new URI(retrieveEndpoint), fileStorageAdapter, irmStorageAdapter, indexDatasetClient);
+        new URI(retrieveEndpoint),
+        fileStorageAdapter,
+        irmStorageAdapter,
+        indexDatasetClient,
+        statusService);
   }
 }
--- a/src/main/java/com/connexta/store/config/TransformConfiguration.java
+++ b/src/main/java/com/connexta/store/config/TransformConfiguration.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.store.config;
+
+import com.connexta.poller.service.StatusService;
+import com.connexta.poller.service.StatusServiceImpl;
+import java.util.concurrent.ExecutorService;
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.task.support.ExecutorServiceAdapter;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+import org.springframework.web.client.RestTemplate;
+
+@Configuration
+public class TransformConfiguration {
+  @Bean
+  public String transformApiVersion(
+      @NotBlank @Value("${endpoints.transform.version}") String transformApiVersion) {
+    return transformApiVersion;
+  }
+
+  @Bean
+  public ExecutorService threadPoolTaskExecutor() {
+    ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+    executor.setCorePoolSize(32);
+    executor.setThreadNamePrefix("polling threadpool");
+    executor.initialize();
+    return new ExecutorServiceAdapter(executor);
+  }
+
+  @Bean
+  public StatusService statusService(@NotNull final ExecutorService executorService) {
+    return new StatusServiceImpl(1, 20, executorService, new RestTemplate());
+  }
+}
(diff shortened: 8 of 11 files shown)


ahoffer:
Its diagram
IMG_20191111_115015746

src/main/java/com/connexta/poller/service/Poller.java
```diff
@@ -0,0 +1,115 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Service;
+
+@Service
+@Slf4j
+public class Poller {
+
+  private final StatusService statusService;
+  private final BlockingQueue<URI> transferQueue;
+  private ExecutorService executor;
```
emmberk: Inject the ExecutorService instead of instantiating it in this class

src/main/java/com/connexta/poller/service/Poller.java
```diff
@@ -0,0 +1,115 @@
...
+  private AtomicBoolean isStarted;
+
+  public Poller(StatusService statusService, BlockingQueue<URI> transferQueue) {
+    this.statusService = statusService;
+    this.transferQueue = transferQueue;
+    isStarted = new AtomicBoolean(false);
+  }
+
+  public void start() {
```
emmberk: I don't think we need these start and stop methods or need to keep track of whether it has been started. I think the Poller should be started when the application is started. Maybe you can implement a finalize method to define how we shutdown the executor.

src/main/java/com/connexta/poller/service/Poller.java
```diff
@@ -0,0 +1,115 @@
...
+      StatusResponse response;
+      try {
+        response = pollResult.get();
+      } catch (InterruptedException | ExecutionException e) {
+        // TODO: Follow-on ticket -- Log the interruption
+        Thread.currentThread().interrupt();
+        return;
+      }
+
+      // TODO Follow-on ticket -- add functionality to call Store Service and other work
```
emmberk: The StoreService should not be called directly here. We should contact the store via REST.

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,109 @@
...
+ * This is class is meant to be instantiated as a Bean, Component, or Service. The class's
+ * responsibilities are to manage the settings for polling and kick off new polling tasks. Other
+ * responsibilities include settings the wait and retry behavior. It also creates and configures the
+ * executor, and ensures that it is shutdown properly.
+ */
+@Service
+@Slf4j
+public class StatusService {
+
+  private final RestTemplate restTemplate;
```
emmberk: We should use `WebClient` instead of `RestTemplate` because `RestTemplate` will be deprecated: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/client/RestTemplate.html

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,109 @@
...
+
+    Task(URI uri) {
+      this.uri = uri;
+    }
+
+    public AttemptResult<StatusResponse> process() {
+      StatusResponse statusResponse = null;
+      try {
+        ResponseEntity<StatusResponse> entity =
+            restTemplate.getForEntity(uri, StatusResponse.class);
```
emmberk: This model should come from the TransformApi models, right?

src/main/java/com/connexta/poller/service/Poller.java
```diff
@@ -0,0 +1,115 @@
...
+      } finally {
+        executor = null;
+        isStarted.set(false);
+      }
+    }
+    log.warn("Stopped WorkerService with {} tasks uncompleted", unfinished.size());
+  }
+
+  @AllArgsConstructor
+  static class Job implements Runnable {
```
emmberk: Shouldn't the logic to send the promote request be in the task in StatusService? Why is this Job needed?

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,109 @@
...
+
+    return builder().polling(new Task(statusUri)).build().start();
+  }
+
+  private PollerBuilder builder() {
+    return PollerBuilder.newBuilder()
+        .withExecutorService(executorService)
+        .stopIfException(false)
+        .withWaitStrategy(WaitStrategies.fixedWait(sleepTime, TimeUnit.SECONDS))
+        .withStopStrategy(StopStrategies.stopAfterDelay(giveUpAfter, TimeUnit.SECONDS));
```
emmberk: 👍 Initially, I thought this was way too low, but I checked with Jeter, and he said that transformation should only take 3-4 seconds, so 20 seconds sounds right to me.

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,109 @@
...
+        // ResourceAccessException thrown if host is not available, unreachable, or not
+        // listening on the port. Exit this attempt, but try again later.
+        continueFor(e);
+      }
+
+      //    If the transform job is done, stop polling
+      if (statusResponse != null && isDone(statusResponse)) {
+
+        // Stop polling
+        return finishWith(statusResponse);
```
emmberk: It would be nice to contain all details about the TransformApi in this one class instead of passing StatusResponse back to the Job in the Poller.

emmberk: Where (in what method) would delete get called on the status endpoint?

src/main/java/com/connexta/poller/service/Poller.java
```diff
@@ -0,0 +1,115 @@
...
+  private void beginWorking() {
+    URI statusUri = null;
+    while (!executor.isTerminated()) {
+      try {
+        statusUri = transferQueue.take();
+      } catch (InterruptedException e) {
+
+        isStarted.set(false);
+        // Reset interrupt
+        Thread.currentThread().isInterrupted();
```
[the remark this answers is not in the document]
ahoffer: I think it is right. It looks weird, but isInterrupted() resets the the interrupt flag on the thread. I don't think I need to re-throw interrupt or interrupt the current thread. Exiting the beginWorking() method should end the thread because it is the bottom of the stack.

src/main/java/com/connexta/poller/service/Poller.java
ahoffer: @connexta/i2o-cdr  If we let the objects call the StatusService directly, we won't even need the queue. That is the approach I'd like to pursue if the team is agreeable to that.

src/main/java/com/connexta/poller/service/StatusService.java
ahoffer: I will switch it to WebClient.

ahoffer: It turns that MockResServiceServer does not work with WebClient the way RestTemplate does. Spring's official advice is to use `OkHttp MockWebServer`.

https://docs.spring.io/spring-framework/docs/current/spring-framework-reference/web-reactive.html#webflux-client-testing

src/main/java/com/connexta/poller/service/Poller.java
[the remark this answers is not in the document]
emmberk: The javadoc says
> The interrupted status of the thread is unaffected by this method.

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Thread.html#isInterrupted()

https://stackoverflow.com/questions/3976344/handling-interruptedexception-in-java says
```java
} catch (InterruptedException e) {
    Thread.currentThread().interrupt();
```

If we don't want to re-interrupt the thread, then I think `Thread.currentThread().isInterrupted();` is meaningless.

src/main/java/com/connexta/poller/service/StatusService.java
emmberk: Right. Our search tests currently use that.

src/main/java/com/connexta/poller/service/StatusService.java
ahoffer: That is my expectation.

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,95 @@
...
+  private final int secondsToLive;
+  @NotNull ExecutorService executorService;
+  @NotNull WebClient webClient;
+
+  public Future<StatusResponse> poll(URI uri) {
+    final HttpStatus expectedHttpStatus = HttpStatus.OK;
+
+    AttemptMaker<StatusResponse> attemptMaker =
+        () -> {
+          StatusResponse statusResponse = null;
```
emmberk: ✏️ Extract the body of this because it will need to be tested independently

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+@Service
+@Slf4j
+@AllArgsConstructor
+public class StatusService {
+
+  private final int secondsBetweenRetries;
+  private final int secondsToLive;
+  @NotNull ExecutorService executorService;
+  @NotNull WebClient webClient;
```
emmberk: Add `private final`
```java
@NotNull private final ExecutorService executorService;
@NotNull private final WebClient webClient;
```

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,95 @@
...
+ * responsibilities include settings the wait and retry behavior. It also creates and configures the
+ * executor, and ensures that it is shutdown properly.
+ */
+@Service
+@Slf4j
+@AllArgsConstructor
+public class StatusService {
+
+  private final int secondsBetweenRetries;
+  private final int secondsToLive;
```
emmberk: ✏️ You can add Min and Max constraints here:
```java
@Min(0) private final int secondsBetweenRetries;
@Min(1) private final int secondsToLive;
```

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,95 @@
...
+import org.springframework.web.reactive.function.client.WebClient;
+import reactor.core.publisher.Mono;
+
+/**
+ * This is class is meant to be instantiated as a Bean, Component, or Service. The class's
+ * responsibilities are to manage the settings for polling and kick off new polling tasks. Other
+ * responsibilities include settings the wait and retry behavior. It also creates and configures the
+ * executor, and ensures that it is shutdown properly.
+ */
+@Service
```
emmberk: ❗️Define the bean in the config classes instead of with `@Service`

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,95 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Service;
+import org.springframework.web.reactive.function.client.WebClient;
+import reactor.core.publisher.Mono;
+
+/**
+ * This is class is meant to be instantiated as a Bean, Component, or Service. The class's
+ * responsibilities are to manage the settings for polling and kick off new polling tasks. Other
+ * responsibilities include settings the wait and retry behavior. It also creates and configures the
+ * executor, and ensures that it is shutdown properly.
```
emmberk: I don't see in this class where the executor is shutdown.

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,95 @@
...
+                        clientResponse ->
+                            Mono.error(
+                                () ->
+                                    new DetailedResponseStatusException(
+                                        clientResponse.statusCode(),
+                                        String.format(
+                                            "Could not get status for %s", uri.toString()))))
+                    .bodyToMono(StatusResponse.class)
+                    .block();
+            // SonarLint said to catch Exception and not Throwable
```
emmberk: ✏️ Delete this comment

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,95 @@
...
+@AllArgsConstructor
+public class StatusService {
+
+  private final int secondsBetweenRetries;
+  private final int secondsToLive;
+  @NotNull ExecutorService executorService;
+  @NotNull WebClient webClient;
+
+  public Future<StatusResponse> poll(URI uri) {
+    final HttpStatus expectedHttpStatus = HttpStatus.OK;
```
emmberk: ✏️ 
```java
private static final HttpStatus EXPECTED_HTTP_STATUS = HttpStatus.OK;
```
or inline

build.gradle
```diff
@@ -13,6 +13,8 @@ buildscript {
         springBootVersion = "2.2.1.RELEASE"
         springfoxSwaggerVersion = "2.9.2"
         storeApiVersion = "0.3.0"
+        //TODO This is a placeholder. This version does not define a status endpoint.
+        transformApiVersion = "0.2.0"
```
emmberk: ✏️ Add a dependency for this (compileOnly) so that dependabot will keep this version up to date

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,95 @@
...
+@Slf4j
+@AllArgsConstructor
+public class StatusService {
+
+  private final int secondsBetweenRetries;
+  private final int secondsToLive;
+  @NotNull ExecutorService executorService;
+  @NotNull WebClient webClient;
+
+  public Future<StatusResponse> poll(URI uri) {
```
emmberk: ✏️ Rename name this method to convey that it kicks of poling instead of blocking. E.g. `submit`

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * This is class is meant to be instantiated as a Bean, Component, or Service. The class's
+ * responsibilities are to manage the settings for polling and kick off new polling tasks. Other
+ * responsibilities include settings the wait and retry behavior. It also creates and configures the
+ * executor, and ensures that it is shutdown properly.
+ */
+@Service
+@Slf4j
+@AllArgsConstructor
+public class StatusService {
```
emmberk: ❗️Extract an interface for the poll method so that it can be injected into StoreService

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
```diff
@@ -102,4 +109,28 @@ public void addIrm(
...
     log.info("Successfully indexed datasetId={}", datasetId);
   }
+
+  /* TODO: Implement the logic call promote/quarantine. Doesn't have to be a lambda. Could be a class the implements Runnable or Callable. */
+  private void poll(URI uri) {
+
+    Future<StatusResponse> pollFuture = statusService.poll(uri);
+
+    // Run on another thread so processing is not blocked
+    executorService.submit(
```
emmberk: ❗️I don't think this should be asynchronous. If submitting the poll fails, we should fail the entire request. I also don't think promote and quarantine should be handled here. Sending the promote/quarantine should happen in the AttemptResult. I think this method can be as simple as. 
```java
private void poll(@NotNull final URI uri) {
  statusService.poll(uri);
}
```

This also simplifies the StatusService#poll method because it shouldn't return anything
```java
public void poll(@NotNull final URI uri) {
```

src/test/java/com/connexta/store/StatusServiceTests.java
```diff
@@ -0,0 +1,73 @@
...
+import okhttp3.HttpUrl;
+import okhttp3.mockwebserver.MockResponse;
+import okhttp3.mockwebserver.MockWebServer;
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.MediaType;
+
+@SpringBootTest
```
emmberk: ❗️`@SpringBootTest` should only be used if you need to start up the whole application. Since this test is only testing the StatusService, this should be a unit test. You shouldn't have to mock things that aren't used in StatusService.

src/test/java/com/connexta/store/StatusServiceTests.java
```diff
@@ -0,0 +1,73 @@
...
+class StatusServiceTests {
```
emmberk: The naming convention is Test not Tests. `StatusServiceTest`

src/main/java/com/connexta/poller/service/StatusService.java
emmberk: ❗️Add tests for this AttemptMaker

src/test/java/com/connexta/store/StatusServiceTests.java
emmberk: ❗️Extract the tests of the AttemptMaker from the tests of the StatusService.

src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1,2 @@
+# Needed to mock final class -- URI
+mock-maker-inline
```
emmberk: ❗️This should not be needed. We shouldn't be mocking URIs.

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
peterhuffer: I think this is okay to be asynchronous. From replications perspective, they don't care if polling submission fails or not, they are not going to be able to fix that error as a client. They care only that we have successfully staged the files requested to be processed. If a user is able to fix it, then they will have to view the failed logs, modify the data and then replication will attempt to stage the data again.

emmberk: It cannot be asynchronous because if we fail to kick off polling, all information will be completely lost because we already returned a success for the ingest request. Additionally, it doesn't make sense to pull the code to call the promote and quarantine endpoints asynchronously from the Attempt because we will need to send the delete request to transform. Let's talk in person to make sure we're on the same page about this.

build.gradle
ahoffer: `compileOnly "com.connexta.transformation:transformation-api-rest-spring-stubs:${transformApiVersion}"`

src/main/java/com/connexta/poller/service/StatusService.java
ahoffer: How would you recommend testing it? Maybe injecting a mock WebClient and using _verify_ to make sure the right methods were called?

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,95 @@
...
+                                        String.format(
+                                            "Could not get status for %s", uri.toString()))))
+                    .bodyToMono(StatusResponse.class)
+                    .block();
+            // SonarLint said to catch Exception and not Throwable
+          } catch (Exception e) {
+            continueFor(e);
+          }
+
+          // TODO: Follow-on ticket -- If the transform job is done, stop polling
```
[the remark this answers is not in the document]
ahoffer: Complete is the name of a status. If transformation is successful and the File can be promoted, then the transform status endpoint  returns complete. However, if the File should be quarantined, the endpoint sets the status to failure.  Complete and failure are both final states. Either one indicates that polling should stop.

src/main/java/com/connexta/poller/service/StatusService.java
emmberk: First I would extract the body of this lambda to a class or method so it can be tested independently. These are the different tests that I would do:
- mock webclient to throw exceptions
- use a real webclient (WebClient.create()) and mock the server (MockWebServer) to
  - return non-202 error codes
  - return invalid response
  - happy path

src/test/java/com/connexta/store/StatusServiceTests.java
ahoffer: I'll switch it over to unit test. It will take a little time. I'll work on it when I get back from my PB.

src/main/java/com/connexta/poller/service/StatusService.java
```diff
@@ -0,0 +1,15 @@
...
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.poller.service;
+
+import java.net.URI;
+
+public interface StatusService {
+
+  @SuppressWarnings("unused")
```
figliold: ❓ Don't we want a result from submit()?

src/main/java/com/connexta/poller/service/StatusServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+      }
+
+      // Job not done, keep polling
+      return justContinue();
+    };
+  }
+
+  private PollerBuilder builder() {
+    return PollerBuilder.newBuilder()
+        .withExecutorService(executorService)
```
figliold: ❓ Do we really want to continue if an exception happens?

src/main/java/com/connexta/poller/service/StatusServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+      status = poll(uri);
+    } catch (ExecutionException | InterruptedException e) {
+      log.info("Polling failed for {}", uri);
+      return;
+    }
+    process(status);
+  }
+
+  private StatusResponse poll(URI uri) throws ExecutionException, InterruptedException {
+    Future<StatusResponse> statusFuture = builder().polling(getAttemptMaker(uri)).build().start();
```
peterhuffer: ❓ Should we be creating and starting a new Poller every poll request?

src/main/java/com/connexta/poller/service/StatusServiceImpl.java
ahoffer: Great question! :grey_question: The answer is yes. If the transform service is down, Spring throws an exception when we make the HTTP request. Also, if polling starts before the status endpoint knows about the job, the endpoint will return a 404 and Spring will turn that response into an exception. Because of situations like that, the application should continue to poll status endpoint because the desired resource might become available. The unit test `testTemporaryFailure` does not succeed is this value is set to `true`.

src/main/java/com/connexta/poller/service/StatusServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+    return statusFuture.get();
+  }
+
+  private AttemptMaker<StatusResponse> getAttemptMaker(URI uri) {
+    return () -> {
+      StatusResponse statusResponse = null;
+      try {
+        statusResponse = restTemplate.getForObject(uri, StatusResponse.class);
+      } catch (ResourceAccessException e) {
+        continueFor(e);
```
mcalcote: ❓ Would we want to return this attempt if this exception occurs?

ahoffer: We do not want to exit polling because of this exception. We want to keep trying until we hit some kind of stop criteria like maximum number of retries, or exceed some time threshold. This exception could be thrown because the transform status endpoint is temporarily unavailable. The service is restarting, or maybe we started polling before the endpoint knew about this particular job.  

As in all things, there is always room for refinement. For instance, if the Exception represents a 400 BAD REQUEST, then it might make sense to call `finishWithException()` instead of `continueFor`. We can tweak and refine as we learn more about how the system actually behaves.

src/main/java/com/connexta/poller/service/StatusServiceImpl.java
```diff
@@ -0,0 +1,96 @@
...
+    } catch (ExecutionException | InterruptedException e) {
+      log.info("Polling failed for {}", uri);
+      return;
+    }
+    process(status);
+  }
+
+  private StatusResponse poll(URI uri) throws ExecutionException, InterruptedException {
+    Future<StatusResponse> statusFuture = builder().polling(getAttemptMaker(uri)).build().start();
+    return statusFuture.get();
```
peterhuffer: ❓ Is this going to block the ingest request? AKA we won't return from the replication request until polling has finished?

ahoffer: Yes, it will block the caller.

ahoffer: Yes. It will block the caller.

src/main/java/com/connexta/poller/service/Poller.java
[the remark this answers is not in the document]
ahoffer: Updated

src/main/java/com/connexta/poller/service/Poller.java
ahoffer: Poller class removed. Executor injected into StatusServiceImpl component.

src/main/java/com/connexta/poller/service/Poller.java
ahoffer: The store service should be invoked via REST.

src/main/java/com/connexta/poller/service/StatusService.java
ahoffer: I tried out WebClient, but swtiched back to RestTemplate. It was easier to unit test ResTemplate and also easier to process Spring HTTP exceptions.

src/main/java/com/connexta/poller/service/Poller.java
ahoffer: Removed the Job class.

PR #229, opened 2019-11-11
Ingest merge into store

mcalcote:
The Ingest capabilities have been merged into the Ion-Store service. 
Local testing is possible with a groovy script that can be run from deployments/local-minio.
The dataset endpoint now returns a 501.

Hero Steps:
Local deployment should be tested and confirm it works as expected


src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
```diff
@@ -6,6 +6,8 @@
  */
 package com.connexta.store.adaptors;
 
+import static com.connexta.store.controllers.StoreController.METACARD_MEDIA_TYPE;
```
mcalcote: The S3MetacardStorageAdapter class was merged into this one.

src/main/java/com/connexta/store/adaptors/StorageAdaptor.java
```diff
@@ -18,6 +18,18 @@
...
+   *
+   * @throws IOException if the file can't be read
+   * @throws StoreException if there is an error when attempting to store
+   */
+  void store(
+      @NotNull @Min(1L) @Max(10737418240L) final Long fileSize,
+      @NotNull final InputStream inputStream,
+      @NotBlank final String key)
+      throws StoreException;
+
```
mcalcote: The MetacardStorageAdapter interface was merged into this one

src/main/java/com/connexta/store/advice/CustomExceptionHandler.java
```diff
@@ -6,7 +6,7 @@
  */
 package com.connexta.store.advice;
 
-import com.connexta.store.exceptions.DetailedErrorAttributes;
+import com.connexta.store.exceptions.common.DetailedErrorAttributes;
```
mcalcote: Moved a few classes around as well

src/main/java/com/connexta/store/clients/TransformClient.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.store.clients;
```
mcalcote: Exact copy from ion-ingest

src/main/java/com/connexta/store/config/TransformClientConfiguration.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.store.config;
```
mcalcote: exact copy from ion-ingest

src/main/java/com/connexta/store/exceptions/IngestException.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.store.exceptions;
```
mcalcote: exact copy from ion-ingest

src/main/java/com/connexta/store/exceptions/TransformException.java
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
```
mcalcote: exact copy from ion-ingest

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
```diff
@@ -40,6 +42,33 @@ public S3StorageAdaptor(@NotNull final AmazonS3 amazonS3, @NotBlank final String
     this.bucket = bucket;
   }
 
+  @Override
+  public void store(
```
emmberk: Combine this store method with the one below. We don't have a need to store differently for metacards, files, and irms.

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
```diff
@@ -132,4 +161,12 @@ private S3Object getS3Object(@NotBlank final String key) throws RetrieveExceptio
 
     return s3Object;
   }
+
+  /** The caller is responsible for closing the returned {@link InputStream}. */
+  @NotNull
+  public InputStream retrieveFileStream(@NotBlank final String key) throws StoreException {
```
emmberk: Delete this method and use `retrieve` instead.

emmberk: This method isn't in the StorageAdaptor interface, so it can never be called

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
```diff
@@ -40,22 +44,8 @@
...
-      @NotNull final InputStream fileInputStream)
-      throws StoreException {
-    final String datasetId = UUID.randomUUID().toString().replace("-", "");
-    fileStorageAdaptor.store(
-        fileSize, mediaType, fileInputStream, datasetId, Map.of(FILE_NAME_METADATA_KEY, fileName));
-    return UriComponentsBuilder.fromUri(retrieveUri)
-        .path(StoreController.RETRIEVE_FILE_URL_TEMPLATE)
-        .build(datasetId);
-  }
+  @NotNull private final S3StorageAdaptor metacardStorageAdaptor;
```
emmberk: `StorageAdaptor` instead of `S3StorageAdaptor`

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -53,41 +61,13 @@
...
-    } catch (IOException e) {
-      throw new ValidationException(
-          String.format(
-              "Unable to read file for createDataset request with mediaType=%s and fileName=%s",
-              mediaType, fileName),
-          e);
-    }
-
-    return ResponseEntity.created(location).build();
+  public Optional<NativeWebRequest> getRequest() {
```
emmberk: Delete this method

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -129,6 +109,72 @@
...
+        inputStream,
+        fileName,
+        metacard.getSize(),
+        metacardInputStream);
+
+    return ResponseEntity.accepted().build();
+  }
+
+  @GetMapping("/{id}")
+  public ResponseEntity<Resource> retrieveMetacard(
```
emmberk: Add annotations to this method, like `retrieveIrm` and `retrieveFile`

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -129,6 +109,72 @@
...
+        file.getContentType(),
+        inputStream,
+        fileName,
+        metacard.getSize(),
+        metacardInputStream);
+
+    return ResponseEntity.accepted().build();
+  }
+
+  @GetMapping("/{id}")
```
emmberk: Should we update this path? E.g. `/tmp/{metacardId}` or `/dataset/{datasetId}/metacard`?

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -129,6 +109,72 @@
...
+  public ResponseEntity<Resource> retrieveMetacard(
+      @ApiParam(required = true) @PathVariable("id") final String id) {
```
emmberk: I think the fact that this id isn't the datasetId is a little confusing. I suggest either updating it to use datasetId and a path or clearly state that this is not the datasetId.

src/main/java/com/connexta/store/service/api/StoreService.java
```diff
@@ -44,6 +45,10 @@ InputStream retrieveIrm(
       @Pattern(regexp = "^[0-9a-zA-Z]+$") @Size(min = 32, max = 32) final String datasetId)
       throws RetrieveException;
 
+  /** The caller is responsible for closing the returned {@link java.io.InputStream}. */
+  @NotNull
+  InputStream retrieveMetacard(@NotBlank final String datasetId) throws RetrieveException;
```
emmberk: ❗️This is not the datasetId

src/main/java/com/connexta/store/service/api/StoreService.java
```diff
@@ -21,13 +21,14 @@
...
       @NotNull @Min(1L) @Max(10737418240L) final Long fileSize,
-      @NotBlank final String mediaType,
+      @NotBlank final String mimeType,
+      @NotNull final InputStream inputStream,
       @NotBlank final String fileName,
-      @NotNull final InputStream fileInputStream)
-      throws StoreException;
+      @NotNull @Min(1L) @Max(10737418240L) final Long metacardFileSize,
+      @NotNull final InputStream metacardInputStream)
+      throws StoreException, TransformException, StoreException;
```
emmberk: `StoreException` is listed twice here

src/main/java/com/connexta/store/exceptions/IngestException.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.connexta.store.exceptions.common.DetailedResponseStatusException;
+import org.springframework.http.HttpStatus;
+
+public class IngestException extends DetailedResponseStatusException {
```
emmberk: Is this ever used? Can we get away with just a TranformException?

src/main/resources/application.yml
```diff
@@ -19,9 +19,12 @@ spring:
 
 # The index endpoint should be deployed in the same network, so we don't need a config.yml file for it.
 endpointUrl:
+  store: http://store:8080/mis/product/
```
emmberk: ❗️ This should not be needed

src/test/java/com/connexta/store/controllers/StoreControllerIngestTests.java
```diff
@@ -0,0 +1,229 @@
...
+  private static final String CORRELATION_ID = "90210";
+  private static final String LAST_MODIFIED_DATE = "1984-04-20T08:08:08Z";
+  private static final MockMultipartFile FILE =
+      new MockMultipartFile(
+          "file", "originalFilename.txt", "text/plain", "file_content".getBytes());
+  private static final MockMultipartFile METACARD =
+      new MockMultipartFile("metacard", "ignored.xml", "application/xml", "content".getBytes());
+
+  @MockBean private StoreService mockStoreService;
+  @Inject private MockMvc mockMvc;
```
emmberk: We should use `WebTestClient` (`@WebFluxTest`) instead of `MockMvc`.

src/test/java/com/connexta/store/controllers/StoreControllerIngestTests.java
```diff
@@ -0,0 +1,229 @@
...
+                .header("Accept-Version", ACCEPT_VERSION))
+        .andExpect(status().is(expectedResponseStatus.value()));
+  }
+
+  private static Stream<Arguments> exceptionThrownByIngestServiceAndExpectedResponseStatus() {
+    return Stream.of(
+        Arguments.of(new StoreException("Test", new Throwable()), HttpStatus.INTERNAL_SERVER_ERROR),
+        Arguments.of(new TransformException(new Throwable()), HttpStatus.INTERNAL_SERVER_ERROR),
+        Arguments.of(
+            new StoreException("Test", new Throwable()), HttpStatus.INTERNAL_SERVER_ERROR));
```
emmberk: This last argument is the same as the first

src/test/java/com/connexta/store/controllers/StoreControllerTest.java
```diff
@@ -39,44 +34,10 @@ public void beforeEach() {
...
-  @NullAndEmptySource
-  @ValueSource(strings = "   ")
-  public void testBadFilename(final String filename, @Mock final MultipartFile mockMultipartFile) {
-    when(mockMultipartFile.getContentType()).thenReturn("testContentType");
-    when(mockMultipartFile.getOriginalFilename()).thenReturn(filename);
-    assertBadRequest(mockMultipartFile);
-  }
-
   @Test
-  void testFileTooLarge(@Mock final MultipartFile mockMultipartFile) {
```
emmberk: Why were these removed? These tests are still valid for the ingest endpoint.

src/test/java/com/connexta/store/controllers/StoreControllerTest.java
```diff
@@ -39,44 +34,10 @@ public void beforeEach() {
...
-  }
-
-  private void assertBadRequest(@Mock MultipartFile mockMultipartFile) {
-    assertThrows(
-        ValidationException.class,
-        () -> storeApi.createDataset(STORE_API_VERSION, mockMultipartFile));
+  void testNotImplementedCreateDataset(@Mock final MultipartFile mockMultipartFile) {
+    assertThat(
+        storeApi.createDataset(STORE_API_VERSION, mockMultipartFile).getStatusCode().value(),
+        is(HttpStatus.SC_NOT_IMPLEMENTED));
```
emmberk: Use the spring http status library
```java
import org.springframework.http.HttpStatus;
...
assertThat(
    storeApi.createDataset(STORE_API_VERSION, mockMultipartFile).getStatusCode(),
    is(HttpStatus.NOT_IMPLEMENTED));
```

src/test/resources/application.yml
```diff
@@ -1,14 +1,19 @@
 server:
-  port: 8080
+  port: 8081
```
emmberk: Why was this changed?

src/test/resources/application.yml
```diff
@@ -1,14 +1,19 @@
...
 endpointUrl:
-  retrieve: http://localhost:${server.port}
+  store: http://localhost:9041/mis/product/
```
emmberk: This should not be needed

src/test/resources/application.yml
```diff
@@ -1,14 +1,19 @@
...
+  retrieve: http://localhost:${server.port}/
+  transform: http://localhost:9090/transform/
```
emmberk: The slash at the end of these url strings should not be needed

src/main/java/com/connexta/store/controllers/StoreController.java
mcalcote: If we want to implement both APIs we have to overwrite this method since they both have it as a default.

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
mcalcote: removed as per your previous comment

src/main/java/com/connexta/store/controllers/StoreController.java
peterhuffer: This will go away with my changes because I am removing OpenAPI spec.

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
mcalcote: What would the mediatype of a metacard be? `text/xml`?

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
mcalcote: that was only to gain access to the retrieveStream convenience method on the impl that is now removed :+1:

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
mcalcote: I see elsewhere we have application xml...

mcalcote: > The registration information for text/xml-external-parsed-entity is
>    in all respects the same as that given for application/xml-external-
>    parsed-entity above (Section 9.3), except that the "Type name" is
>    "text".
https://tools.ietf.org/html/rfc7303

emmberk: I think it's always application/xml. You can debug sending a postman request and double check though. Or just guess application/xml and we can fix it later.

src/main/java/com/connexta/store/controllers/StoreController.java
mcalcote: added `public static final String RETRIEVE_METACARD_URL_TEMPLATE = "/dataset/{datasetId}/metacard";`

src/main/java/com/connexta/store/controllers/StoreController.java
mcalcote: Updated it to datasetId, hopefully it is more consistent/clear now

src/main/java/com/connexta/store/exceptions/IngestException.java
mcalcote: Only for transform exception to extend 🤔 I will remove

src/test/java/com/connexta/store/controllers/StoreControllerIngestTests.java
mcalcote: 🗑

src/test/java/com/connexta/store/controllers/StoreControllerTest.java
mcalcote: Didn't even see that 😂  👍

src/test/resources/application.yml
mcalcote: Umm... I think there is something wrong with the test because when you run them from the terminal there was a port collision, haven't investigated yet

src/test/resources/application.yml
mcalcote: at least at one point, the retrieve endpoint was responding without it so we would see results like `http://localhost:808005b04fc7bd0a497b9e946a75aeec293b` Haven't investigated yet

src/main/java/com/connexta/store/service/api/StoreService.java
mcalcote: Updated so the metacard is ingested with the datasetId

src/test/java/com/connexta/store/IngestITests.java
```diff
@@ -0,0 +1,356 @@
...
+    storeMockRestServiceServer.reset();
+    transformMockRestServiceServer.verify();
+    transformMockRestServiceServer.reset();
+
+    amazonS3.listObjects(s3Bucket).getObjectSummaries().stream()
+        .map(S3ObjectSummary::getKey)
+        .forEach(key -> amazonS3.deleteObject(s3Bucket, key));
+    amazonS3.deleteBucket(s3Bucket);
+  }
+  //
```
ahoffer: If you don't want the tests to run, but you don't want to delete them either, use the `@Disabled` annotation from Junit 5.

mcalcote: Good suggestion, thanks for the reminder

src/test/java/com/connexta/store/controllers/StoreControllerTest.java
mcalcote: temporary solution only

mcalcote: I put the same checks in the storecontrolleringesttest

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -129,6 +109,111 @@
...
+        file.getContentType(),
+        inputStream,
+        fileName,
+        metacard.getSize(),
+        metacardInputStream);
+
+    return ResponseEntity.accepted().build();
+  }
+
+  @ApiOperation(
```
peterhuffer: Not as part of this PR, but these should all be defined in the API repo, right?

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -129,6 +109,111 @@
     return ok().build();
   }
 
+  @Override
+  public ResponseEntity<Void> ingest(
+      String acceptVersion,
+      @DateTimeFormat(iso = DateTimeFormat.ISO.DATE_TIME) OffsetDateTime lastModified,
+      MultipartFile file,
+      String correlationId,
+      MultipartFile metacard) {
```
figliold: ❓ Could you use StringUtils.isBlank() here?

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -129,6 +109,111 @@
...
+    if (lastModified == null || lastModified.toString().isBlank()) {
+      throw new ServerWebInputException(
+          String.format("%s is missing or blank", LAST_MODIFIED)); // TODO Replace this exception
+    }
+    String fileName = file.getOriginalFilename();
+    log.info("Ingest request received fileName={}", fileName);
```
figliold: ❓ Should these input streams get closed?

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -129,6 +109,111 @@
...
+      value = RETRIEVE_METACARD_URL_TEMPLATE,
+      produces = {MediaType.APPLICATION_OCTET_STREAM_VALUE, MediaType.APPLICATION_XML_VALUE},
+      method = RequestMethod.GET)
+  public ResponseEntity<Resource> retrieveMetacard(
+      @Pattern(regexp = "^[0-9a-zA-Z]+$")
+          @Size(min = 32, max = 32)
+          @ApiParam(value = "The ID of the dataset. ", required = true)
+          @PathVariable("datasetId")
+          final String datasetId) {
+    InputStream inputStream = null;
```
figliold: ❓ Can you use a try-with-resources here?

src/main/java/com/connexta/store/controllers/StoreController.java
emmberk: I don't think they need to be defined in the API. They would never be versioned and no client should ever construct the URI themselves.

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -129,6 +109,111 @@
...
+          inputStream.close();
+        } catch (IOException ioe) {
+          log.warn(
+              "Unable to close InputStream when retrieving metacard from datasetId={}",
+              datasetId,
+              ioe);
+        }
+      }
+
+      log.warn("Unable to retrieve metacard datasetId={}", datasetId, e);
```
figliold: ❓ Should the datasetId be in the exception message?

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -129,6 +109,111 @@
...
+      throw new StoreException(String.format("Unable to retrieve metacard: %s", e.getMessage()), e);
```
figliold: ❓ Should we really be catching Throwable here?

src/main/java/com/connexta/store/controllers/StoreController.java
peterhuffer: Why would a client need to construct the URL? Wouldn't they just say give me the metacard (should be more general in the future) for a given dataset?

emmberk: Our API doesn't work that way currently. No client ever constructs URLs, except clients of the ingest endpoint. Additionally, the retrieve endpoints are URIs, not URLs. They should never be versioned or have headers/params.

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
```diff
@@ -42,18 +42,22 @@ public S3StorageAdaptor(@NotNull final AmazonS3 amazonS3, @NotBlank final String
...
     // TODO check if id already exists
 
     final ObjectMetadata objectMetadata = new ObjectMetadata();
-    objectMetadata.setContentType(mediaType);
-    objectMetadata.setContentLength(size);
-    objectMetadata.setUserMetadata(metadata);
+    if (!mediaType.isEmpty()) {
+      objectMetadata.setContentType(mediaType);
+    }
+    if (!mediaType.isEmpty()) {
```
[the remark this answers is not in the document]
mcalcote: Typo, should have been `metadata`

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
```diff
@@ -42,18 +42,22 @@ public S3StorageAdaptor(@NotNull final AmazonS3 amazonS3, @NotBlank final String
...
+      Map<String, String> metadata)
       throws StoreException {
     // TODO check if id already exists
 
     final ObjectMetadata objectMetadata = new ObjectMetadata();
-    objectMetadata.setContentType(mediaType);
-    objectMetadata.setContentLength(size);
-    objectMetadata.setUserMetadata(metadata);
+    if (!mediaType.isEmpty()) {
+      objectMetadata.setContentType(mediaType);
```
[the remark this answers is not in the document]
mcalcote: We were already doing this in the case of metacards: https://github.com/connexta/ion-ingest/blob/master/src/main/java/com/connexta/ingest/adaptors/MetacardStorageAdaptor.java

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
```diff
@@ -42,18 +42,22 @@ public S3StorageAdaptor(@NotNull final AmazonS3 amazonS3, @NotBlank final String
...
-      @NotNull Map<String, String> metadata)
+      Map<String, String> metadata)
       throws StoreException {
     // TODO check if id already exists
 
     final ObjectMetadata objectMetadata = new ObjectMetadata();
-    objectMetadata.setContentType(mediaType);
-    objectMetadata.setContentLength(size);
-    objectMetadata.setUserMetadata(metadata);
+    if (!mediaType.isEmpty()) {
```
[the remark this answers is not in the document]
mcalcote: Users may intentionally be storing without a mediaType or metadata. I was trying to avoid setting nulls.

src/main/java/com/connexta/store/adaptors/StorageAdaptor.java
```diff
@@ -26,10 +26,10 @@
    */
   void store(
       @NotNull @Min(1L) @Max(10737418240L) final Long fileSize,
-      @NotBlank final String mediaType,
+      final String mediaType,
```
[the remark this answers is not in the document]
mcalcote: Yes, but that may not be the final answer. In the case of metacards no mediatype was previously being specified, but I did update those to `MediaType.APPLICATION_XML`.

src/main/java/com/connexta/store/adaptors/StorageAdaptor.java
```diff
@@ -26,10 +26,10 @@
...
       @NotNull final InputStream inputStream,
       @NotBlank final String key,
-      @NotNull Map<String, String> metadata)
+      Map<String, String> metadata)
```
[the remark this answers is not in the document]
mcalcote: Yes

src/main/java/com/connexta/store/controllers/StoreController.java
mcalcote: Added close()

src/main/java/com/connexta/store/controllers/StoreController.java
mcalcote: It makes things a bit mess, will save for a rainy day.

src/main/java/com/connexta/store/controllers/StoreController.java
mcalcote: Life is wild

emmberk: Yes. We have to catch throwable to make sure that the input stream gets closed. I don't know of another solution.

emmberk:
I heroed the happy path

src/test/java/com/connexta/store/StoreITests.java
```diff
@@ -528,4 +470,124 @@ protected boolean matchesSafely(URI uri) {
       }
     };
   }
+
+  @Test
+  public void testSuccessfulIngestRequest() throws Exception {
+    AtomicReference<String> metacardId = new AtomicReference<>();
```
ahoffer: It looks like this var refers to a URI/URL. Calling it `metacardUri` or similar would help readers know what to expect.

src/test/java/com/connexta/store/StoreITests.java
```diff
@@ -528,4 +470,124 @@ protected boolean matchesSafely(URI uri) {
...
+        .header(
+            "Accept-Version",
+            "0.5.0") // TODO inject this like we do for the transformApiVersion in TransformClient
+        .header(LAST_MODIFIED, LAST_MODIFIED_DATE)
+        .exchange()
+        .expectStatus()
+        .isAccepted();
+
+    String id[] = StringUtils.split(metacardId.get(), '/');
+    assertNotNull(amazonS3.getObject(metacardBucket, id[3]));
```
ahoffer: What does `id[3]` represent? Consider assigning `id[3]` to a local var whose name shows what it represents.

src/test/java/com/connexta/store/StoreITests.java
```diff
@@ -8,56 +8,68 @@
...
 import static com.connexta.store.controllers.StoreController.CREATE_DATASET_URL_TEMPLATE;
+import static com.connexta.store.controllers.StoreController.METACARD_MEDIA_TYPE;
 import static com.connexta.store.controllers.StoreController.SUPPORTED_METADATA_TYPE;
-import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.not;
-import static org.hamcrest.core.AllOf.allOf;
-import static org.hamcrest.core.IsNull.notNullValue;
+import static org.junit.jupiter.api.Assertions.*;
```
jhunzik: ✏️ Only import what we need.

src/test/java/com/connexta/store/controllers/StoreControllerIngestTests.java
```diff
@@ -0,0 +1,334 @@
...
+                CORRELATION_ID,
+                METACARD)
+            .getStatusCode(),
+        is(HttpStatus.ACCEPTED));
+  }
+
+  @ParameterizedTest
+  @NullAndEmptySource
+  @ValueSource(strings = "   ")
+  public void testBadFilename(final String filename) {
```
jhunzik: ❓ Why would these situations return 202 Accepted? Wouldn't we be returning something along the lines of 5XX?

src/test/java/com/connexta/store/controllers/StoreControllerIngestTests.java
```diff
@@ -0,0 +1,334 @@
...
+                METACARD)
+            .getStatusCode(),
+        is(HttpStatus.ACCEPTED));
+  }
+
+  @Test
+  void testIngestFileTooLarge() {
+    when(mockFile.getContentType()).thenReturn("testContentType");
+    when(mockFile.getOriginalFilename()).thenReturn("testFilename");
+    when(mockFile.getSize()).thenReturn(MAX_FILE_BYTES + 1);
```
jhunzik: I don't think think this tests what will actually happen. The file size validation will happen on the Spring side before our controller.

src/test/java/com/connexta/store/StoreITests.java
mcalcote: Updated to `datasetUri` 👍

src/test/java/com/connexta/store/controllers/StoreControllerIngestTests.java
ahoffer: I think it would request a BAD REQUEST code because the filename in the request is not valid.

src/test/java/com/connexta/store/StoreITests.java
mcalcote: 🤦‍♂

src/test/java/com/connexta/store/StoreITests.java
emmberk: We should use `datasetId`. https://github.com/connexta/ion-transformation-api/pull/68#discussion_r345861497

src/test/java/com/connexta/store/StoreITests.java
mcalcote: I bumped up my settings in intellij -> shouldn't happen again

src/test/java/com/connexta/store/controllers/StoreControllerIngestTests.java
jhunzik: Yeah, you're right. To clarify, my comment refers to all of the bad parameter tests at the bottom of the test class. It looks like they all assert Accepted.

mcalcote:
Hero'd on 80fdb69
Confirmed ingest, adding irm, retrieving product/metacard/irm, and interacting with the create dataset endpoint.

PR #234, opened 2019-11-20
Add setStatus to StorageAdaptor

jhunzik:
**[Ready for Review]**
This PR adds a setStatus capability to the StorageAdaptor interface that will allow it to tag S3 objects. It also replaces Minio in the tests and local deployment with [Localstack](https://github.com/localstack/localstack)

This change should be tested by running the deploy.groovy script in the <project-root>/deployments/local directory and verifying that files can still be ingested and retrieved.


src/main/java/com/connexta/store/service/api/StoreService.java
```diff
@@ -54,4 +54,6 @@ void addIrm(
       @NotNull @Min(1L) @Max(10737418240L) final long fileSize,
       @Pattern(regexp = "^[0-9a-zA-Z]+$") @Size(min = 32, max = 32) final String datasetId)
       throws IndexDatasetException;
+
+  void promote(@NotNull String datasetId);
```
emmberk: ✏️ I think we should use "unstage" instead of "promote"

src/main/java/com/connexta/store/adaptors/StoreStatus.java
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (c) 2019 Connexta, LLC
+ *
+ * Released under the GNU Lesser General Public License version 3; see
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.store.adaptors;
+
+public class StoreStatus {
```
emmberk: ✏️ Enum

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
```diff
@@ -110,6 +121,14 @@ public StorageAdaptorRetrieveResponse retrieve(@NotBlank final String key)
     }
   }
 
+  @Override
+  public void updateStatus(String datasetId, String storeStatus) {
```
emmberk: I think we should protect against setting any type of status, because only a few should be allowed.

src/main/java/com/connexta/store/adaptors/StorageAdaptor.java
```diff
@@ -39,4 +39,12 @@ void store(
...
   @NotNull
   StorageAdaptorRetrieveResponse retrieve(@NotBlank final String key) throws RetrieveException;
+
+  /**
+   * Updates the {@link StoreStatus} of {@code dataset} for a given {@code datasetId}
+   *
+   * @param datasetId the Id of a dataset
+   * @param storeStatus the new {@link StoreStatus} for a dataset
+   */
+  void updateStatus(String datasetId, String storeStatus);
```
peterhuffer: ❓ Does this have to be called after `store()`?

jhunzik: No, store adds the tag in its request

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
jhunzik: That did cross my mind. I'll look into changing it.

src/main/java/com/connexta/store/adaptors/StorageAdaptor.java
peterhuffer: I like the direction of this. I think StoreStatus can go away and maybe just pass in a parameter called `String group`, and let the client group Datasets arbitrarily.

src/main/java/com/connexta/store/service/api/StoreService.java
jhunzik: I like

src/main/java/com/connexta/store/adaptors/StorageAdaptor.java
peterhuffer: Ah I see now.

src/main/java/com/connexta/store/adaptors/StoreStatus.java
peterhuffer: I like the string. It keeps the Adapter non specific to Ion needs.

emmberk: The Adaptor shouldn't allow every string though. Either we limit it to an enum, or it should reject statuses that aren't one of the ones listed here.

peterhuffer: Only the Store cares what we tag with, why does the Adapter care?

src/main/java/com/connexta/store/adaptors/StorageAdaptor.java
jhunzik: I made that class to make tagging more convenient and less error prone, not necessarily to restrict what a client can tag.

peterhuffer: I guess I'm getting at I think these constants belong only to the Store, not the Adapter

jhunzik: Hmm, at that point should we just change this method to be `updateTag(String datasetId, String key, String value)` or a Map instead of one k-v pair.

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
```diff
@@ -92,6 +92,9 @@ public void addIrm(
     log.info("Successfully indexed datasetId={}", datasetId);
   }
 
+  @Override
+  public void unstage(String datasetId) {}
```
emmberk: ✏️ Since this method isn't implemented, could you remove it? Or throw an exception?

src/main/java/com/connexta/store/adaptors/StorageAdaptor.java
emmberk: ❗️Add validation constraints on datasetId and storeStatus (and tests)

src/main/java/com/connexta/store/adaptors/StoreStatus.java
emmberk: ✏️ The only one who should ever use these strings is the class that calls updateStatus. I suggest moving these to that class.

emmberk: Let's discuss in person.

src/test/java/com/connexta/store/adaptors/S3StorageAdaptorITests.java
```diff
@@ -153,11 +157,36 @@ public void testStoreWithContentLengthNotMatching() {
...
               MediaType.APPLICATION_XML_VALUE,
               new ByteArrayInputStream(ASDF.getBytes()),
-              KEY,
+              DATASET_ID,
               Map.of());
         });
   }
 
+  @Test
+  void testUpdatingStoreStatus() {
```
emmberk: ❗️I'd expect a test for updating a status for a datasetId that does not yet exist

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
```diff
@@ -110,6 +121,14 @@ public StorageAdaptorRetrieveResponse retrieve(@NotBlank final String key)
...
+    final Tag updatedTag = new Tag(STATUS_KEY, storeStatus);
+    final ObjectTagging objectTagging = new ObjectTagging(ImmutableList.of(updatedTag));
+
+    amazonS3.setObjectTagging(new SetObjectTaggingRequest(bucket, datasetId, objectTagging));
```
emmberk: ❗️Can this fail? What are the situations that would cause this to fail? I think that should be documented in the javadoc and tested.

jhunzik: Yeah it can fail for the same reasons as other instances of AmazonS3: the client is unavailable, the S3 service is unavailable, etc.

src/main/java/com/connexta/store/adaptors/StorageAdaptor.java
jhunzik: I will, but I'm still not convinced that they do anything. (If you're referring to annotations).

emmberk: If annotations work, that's good. But we should protect against null and empty values at least.

peterhuffer: ❗️ What happens if the datasetId does not exist? What happens if there is an error reading the file? Client needs to be able to respond to these cases.

jhunzik: I'll add a check for the datasetId. The error reading the file would be an exception internal to S3. What's weird is that the javadoc for setObjectTagging doesn't say it throws anything. https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3.html#setObjectTagging-com.amazonaws.services.s3.model.SetObjectTaggingRequest-

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
```diff
@@ -110,18 +121,25 @@ public StorageAdaptorRetrieveResponse retrieve(@NotBlank final String key)
     }
   }
 
+  @Override
+  public void updateStatus(@NotBlank String key, @NotBlank String storeStatus) {
+    final Tag updatedTag = new Tag(STATUS_KEY, storeStatus);
```
emmberk: ✏️ Shouldn't the STATUS_KEY constant be moved to this class because it's only used here?

src/test/java/com/connexta/store/adaptors/S3StorageAdaptorITests.java
```diff
@@ -153,11 +157,42 @@ public void testStoreWithContentLengthNotMatching() {
...
+        Map.of());
+    assertEquals(STAGED, getStatusTag(DATASET_ID));
+
+    // Promote file to stored status
+    storageAdaptor.updateStatus(DATASET_ID, STORED);
+    assertEquals(STORED, getStatusTag(DATASET_ID));
+  }
+
+  @Test
+  void testUpdatingStoreStatusInvalidDatasetId() {
```
emmberk: ✏️ Could parameterize this to test all of the status values

src/test/java/com/connexta/store/adaptors/S3StorageAdaptorITests.java
```diff
@@ -153,11 +157,42 @@ public void testStoreWithContentLengthNotMatching() {
...
+  @Test
+  void testUpdatingStoreStatus() {
+    // Store a file and set status as staged
+    storageAdaptor.store(
+        4L,
+        MediaType.APPLICATION_XML_VALUE,
+        new ByteArrayInputStream(ASDF.getBytes()),
+        DATASET_ID,
+        Map.of());
+    assertEquals(STAGED, getStatusTag(DATASET_ID));
```
emmberk: ✏️ Unit test methods should only test one thing. I suggest moving this assert to `testSuccessfulStoreRequest` so that this method only asserts the new status.

src/test/java/com/connexta/store/StoreITests.java
```diff
@@ -83,10 +82,11 @@
...
 
   private static final String DATASET_ID = "341d6c1ce5e0403a99fe86edaed66eea";
-  private static final String MINIO_ADMIN_ACCESS_KEY = "admin";
-  private static final String MINIO_ADMIN_SECRET_KEY = "12345678";
   private static final String LAST_MODIFIED_DATE = "2017-06-11T14:32:28Z";
-  private static final int MINIO_PORT = 9000;
+  public static final String ACCESS_KEY = "access-key";
+  public static final String SECRET_KEY = "secret-key";
+  private static final String DOCKER_IMAGE_NAME = "localstack/localstack:0.10.5";
+  private static final int LOCALSTACK_PORT = 4572;
```
emmberk: ✏️ Can these all be inlined? They should only ever be used once in one place.

src/test/java/com/connexta/store/adaptors/S3StorageAdaptorITests.java
```diff
@@ -153,11 +157,42 @@ public void testStoreWithContentLengthNotMatching() {
...
+
+    // Promote file to stored status
+    storageAdaptor.updateStatus(DATASET_ID, STORED);
+    assertEquals(STORED, getStatusTag(DATASET_ID));
```
emmberk: ✏️ Should we test that the file is still retrievable after changing the status?

src/test/java/com/connexta/store/adaptors/S3StorageAdaptorITests.java
```diff
@@ -6,23 +6,28 @@
  */
 package com.connexta.store.adaptors;
 
+import static com.connexta.store.adaptors.StoreStatus.STAGED;
+import static com.connexta.store.adaptors.StoreStatus.STATUS_KEY;
+import static com.connexta.store.adaptors.StoreStatus.STORED;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
```
emmberk: ✏️ I think `assertThat` is preferred over `assertEquals` and `assertTrue` because it has better failure messages.

deployments/local/scripts/create_buckets.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
```
figliold: ❓ Do we need to create a Windows version of this script?

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
```diff
@@ -110,18 +121,25 @@ public StorageAdaptorRetrieveResponse retrieve(@NotBlank final String key)
     }
   }
 
+  @Override
+  public void updateStatus(@NotBlank String key, @NotBlank String status) {
```
figliold: ❓ Can these be moved inside the if s3ObjectExists check?

src/test/java/com/connexta/store/adaptors/S3StorageAdaptorITests.java
jhunzik: I think that kind of test is better suited for the StoreService where the logic to control what can and cannot be retrieved will live.

src/test/java/com/connexta/store/adaptors/S3StorageAdaptorITests.java
jhunzik: I'll update them.

deployments/local/scripts/create_buckets.sh
jhunzik: No, this script runs in the docker container.

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
jhunzik: Sure, I'll move them in

deployments/local/scripts/create_buckets.sh
jhunzik: No, this runs in the Docker container.

src/test/java/com/connexta/store/adaptors/S3StorageAdaptorITests.java
emmberk: Updating status should not change the file that was stored. We don't currently have any tests for that.

deployments/local/README.md
```diff
@@ -1,9 +1,11 @@
...
+# Local Deployment
 
 ## Description
-This is an example local deployment of the Store Service that also starts up a local Minio server
-on port 9000, creates an ingest-quarantine bucket, and uses that for the S3 store. It also includes
-a deploy.sh script that will attempt to create the docker network needed.
+This is an example local deployment of the Store Service that also starts up a Localstack Docker image that provides
+an implementation of the S3 API on port 4572. It also creates 3 S3 buckets: ingest-quarantine, metacard-quarantine,
+and irm. A basic dashboard UI is available on port 8055 and a JSON record of all requests to S3 will be stored in
+<project-root>/.localstack/data/s3_api_calls.json .
```
mcalcote: ✏️ Please mention that the `.localstack` directory needs to be created beforehand

jhunzik:
Hero'd again with MacOS on commit 10fa3e12ad301e7219f2781494b84ff58a91dec0. Was able to create the temp localstack data directory with the deploy script, deploy store and localstack, ingest a text file, and retrieve the file by datasetId.

deployments/local/scripts/deploy.groovy
```diff
@@ -23,7 +23,17 @@ def header(message) {
     println ""
 }
 
+def createLocalS3Store() {
+    header("Deleting contents of .localstack")
+    def tmpDir = new File("../../.localstack")
```
mcalcote: `    def tmpDir = new File("../../../.localstack")
`

deployments/local/scripts/deploy.groovy
```diff
@@ -23,7 +23,17 @@ def header(message) {
...
+        tmpDir.deleteDir()
+    }
+    tmpDir.mkdir()
+}
+
 header("Deploying the application on " + STACK)
+createLocalS3Store()
 run("docker stack rm " + STACK)
 run("docker network create --driver=overlay --attachable " + NETWORK)
 run("docker stack deploy -c ../../docker-compose.yml -c docker-override.yml " + STACK)
```
mcalcote: `run("docker stack deploy -c ../../../docker-compose.yml -c ../docker-override.yml " + STACK)
`

jhunzik: I fixed this in my last force push. Now we run a check to make sure you run the script in `scripts` directory

PR #235, opened 2019-11-21
Add support for Transform Status polling

peterhuffer:
- Upgrades transform API version and updates transform request and adds transform status poller
- Upgrades store API version
- Modifies controller and StoreService APIs to handle multiple Datas
- Update code to deal with 36 length UUIDs

Correct error handling is currently difficult because I don't know if Datasets or Data exist or not based on the current StorageAdapter API.

To hero:
- Send an ingest request.
- Verify the IRM, file, and metacard can be downloaded


src/main/java/com/connexta/store/config/StoreControllerConfiguration.java
```diff
@@ -10,13 +10,23 @@
...
 public class StoreControllerConfiguration {
 
+  private static final long TEN_GIGABYTES = 10737418240L;
+
   @Bean
   public String storeApiVersion(
       @NotBlank @Value("${endpoints.store.version}") String storeApiVersion) {
     return storeApiVersion;
   }
+
```
jhunzik: ❓ Shouldn't this already be set by a property in the application.yml? https://github.com/connexta/ion-store/blob/e007a56b48aa6476955357372496ba9061fa6b18/src/main/resources/application.yml#L17

src/main/java/com/connexta/store/service/api/StoreService.java
```diff
@@ -6,52 +6,48 @@
...
+/** Provides operations to take on Datasets. */
 public interface StoreService {
 
+  /**
+   * Ingests a file and metacard.
+   *
+   * @param fileSize the size of the file
+   * @param mimeType the type of the file
+   * @param inputStream the contents of the file
+   * @param fileName the files name
```
jhunzik: ✏️ `file's`

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
```diff
@@ -6,128 +6,208 @@
...
-import javax.validation.constraints.Max;
-import javax.validation.constraints.Min;
+import java.util.concurrent.BlockingQueue;
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
-import javax.validation.constraints.Pattern;
-import javax.validation.constraints.Size;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.core.io.Resource;
```
jhunzik: I think we should avoid Spring-specific code in our business logic. The response should be handled in the controller.

src/main/java/com/connexta/store/config/StoreControllerConfiguration.java
peterhuffer: Yeah I removed this guy.

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
peterhuffer: Yeah we are going to have to come in here and rip out all the spring code after we re-define our interfaces.

src/main/java/com/connexta/store/poller/TransformStatusPoller.java
```diff
@@ -0,0 +1,281 @@
...
+          if (transformationPollResponse.getBody() == null) {
+            log.debug(
+                "No body available in the successful transform status response for status {}. This URL will be retried.",
+                transformPollUrl);
+            return true;
+          }
+
+          return handlePollResponse(datasetId, transformationPollResponse.getBody());
+        } else {
+          log.debug(
```
mcalcote: ✏️ It looks like this log could be combined with the one above
```        
        if (transformationPollResponse != null && transformationPollResponse.getBody() != null) {
            return handlePollResponse(datasetId, transformationPollResponse.getBody());
          }
        log.debug(
                "No body available in the successful transform status response for status {}. This URL will be retried.", 
                transformPollUrl);
        return true;

File: src/main/java/com/connexta/store/poller/TransformStatusPoller.java [hunk 5]
```diff
@@ -0,0 +1,281 @@
...
+              "No body available in the successful transform status response for status {}. This URL will be retried.",
+              transformPollUrl);
+          return true;
+        }
+      } else if (HttpStatus.NOT_FOUND.equals(httpStatus)) {
+        log.debug(
+            "Transform URL status for {} does not exist anymore, removing from task queue.",
+            transformPollUrl);
```
mcalcote: ❓ When would we run into this scenario? Would we want to log something more severe if we were trying to poll something not there?

src/main/java/com/connexta/store/poller/TransformStatusPoller.java
```diff
@@ -0,0 +1,281 @@
...
+            .exchange()
+            .block();
+
+    if (response != null) {
+      final HttpStatus httpStatus = response.statusCode();
+      if (HttpStatus.OK.equals(httpStatus)) {
+        deleteTransformationStatus(datasetId);
+        return false;
+      } else if (HttpStatus.NOT_FOUND.equals(httpStatus)) {
+        log.debug("Failed to add metadata to dataset {} because it no longer exists.", datasetId);
```
mcalcote: ✏️ This seems like another of those catastrophic cases where a stronger log may be helpful

src/main/java/com/connexta/store/poller/TransformStatusPoller.java
```diff
@@ -0,0 +1,281 @@
...
+                UriComponentsBuilder.fromPath(
+                        TransformConfiguration.TRANSFORM_STATUS_DELETE_TEMPLATE)
+                    .build(datasetId)
+                    .toASCIIString())
+            .exchange()
+            .block();
+
+    if (response != null) {
+      final HttpStatus httpStatus = response.statusCode();
+      if (HttpStatus.NO_CONTENT.equals(httpStatus)) {
```
mcalcote: ❓ Any reason to have all this logic if this is a best effort scenario with no retries?

src/main/java/com/connexta/store/clients/TransformClient.java
```diff
@@ -6,44 +6,78 @@
...
+            .block();
+
+    if (response != null) {
+      HttpStatus httpStatus = response.statusCode();
+      if (HttpStatus.CREATED.equals(httpStatus)) {
+        return extractLocation(response);
+      }
+
+      throw new DetailedResponseStatusException(
+          HttpStatus.INTERNAL_SERVER_ERROR, "Unsuccessful transform request sent.");
```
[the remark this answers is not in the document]
peterhuffer: I will log it. Still need to go through some of these classes and add logs. I don't think we want to report another service response's status code to our client though. I figure that in the case transform keeps failing, the debugging would occur in the transform service.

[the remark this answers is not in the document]
peterhuffer: Eventually these errors will not be sent to the client. As soon as ingest receives the files, it is successful from a client's perspective. If we fail to send the transform request, then it is an error with our internal services that need to be fixed.

src/main/java/com/connexta/store/poller/TransformStatusPoller.java
```diff
@@ -0,0 +1,281 @@
...
+            "Timed out trying to resubmit transform status {} for dataset {}. The file corresponding to this dataset should be reingsted.",
+            transformStatusTask.getTransformStatusUrl(),
+            transformStatusTask.getDatasetId());
+      }
+    }
+  }
+
+  /** @return {@code true} if the task should be retried, otherwise {@code false} */
+  private boolean handleTask(TransformStatusTask transformStatusTask) {
+    final String transformPollUrl = transformStatusTask.getTransformStatusUrl().toString();
```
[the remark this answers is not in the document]
peterhuffer: The `WebClient.post().uri()` function takes a string.

src/main/java/com/connexta/store/poller/TransformStatusPoller.java
peterhuffer: Whoops, that second log message should be different. Bad copy paste.

src/main/java/com/connexta/store/poller/TransformStatusPoller.java
peterhuffer: The only case I could think right now is if someone happen to manually delete a dataset that had not finished polling. I logged it at info.

src/main/java/com/connexta/store/poller/TransformStatusPoller.java
peterhuffer: Mostly for logging, but I could be convinced to delete it cause I decided not to test these cases since they are pretty untestable.

peterhuffer: And transform should be logging all these cases.

peterhuffer: Yeah i'll just remove all this logic and log the attempt.

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -37,75 +37,46 @@
...
-  public static final String RETRIEVE_METACARD_URL_TEMPLATE = "/dataset/{datasetId}/metacard";
-  public static final MediaType IRM_MEDIA_TYPE = new MediaType("application", "dni-tdf+xml");
+  public static final String RETRIEVE_DATA_URL_TEMPLATE = "/dataset/{datasetId}/{dataType}";
+  public static final String QUARANTINE_URL_TEMPALTE = "/dataset/{datasetId}/quarantine";
+  public static final String ADD_METADATA_URL_TEMPLATE = "/dataset/{datasetId}";
   public static final String IRM_MEDIA_TYPE_VALUE = "application/dni-tdf+xml";
-  public static final MediaType METACARD_MEDIA_TYPE = MediaType.APPLICATION_XML;
-  @NotNull private final StoreService storeService;
-  @NotBlank private final String storeApiVersion;
+  private static final long TEN_GIGABYTES = 1073741824L;
```
jhunzik: ✏️ This isn't used anywhere

src/main/java/com/connexta/store/controllers/StoreController.java
```diff
@@ -142,200 +131,73 @@
...
             response = ErrorMessage.class),
         @ApiResponse(
             code = 501,
-            message = "The requested API version is not supported and therefore not implemented. ",
+            message = "The requested API version is not supported and therefore not implemented.",
             response = ErrorMessage.class)
       })
   @RequestMapping(
-      value = RETRIEVE_FILE_URL_TEMPLATE,
+      value = "/dataset/{datasetId}/{dataType}",
```
jhunzik: ✏️ Why not use the constant you defined?

src/main/java/com/connexta/store/poller/TransformStatusPoller.java
```diff
@@ -0,0 +1,270 @@
...
+    if (shouldRetry) {
+      boolean resubmitted = transformStatusQueue.offer(transformStatusTask, 10, TimeUnit.SECONDS);
+      if (resubmitted) {
+        log.debug(
+            "Transform status {} will be tried again for dataset {}.",
+            transformStatusTask.getTransformStatusUrl(),
+            transformStatusTask.getDatasetId());
+      } else {
+        log.debug(
+            "Timed out trying to resubmit transform status {} for dataset {}. The file corresponding to this dataset should be reingsted.",
```
jhunzik: ✏️ `re-ingested`

src/main/java/com/connexta/store/poller/TransformStatusPoller.java
```diff
@@ -0,0 +1,270 @@
...
+    this.storeWebClient = storeWebClient;
+  }
+
+  // TODO 11/20/2019 PeterHuffer: This class should be managed and be able to be properly destroyed
+  @SuppressWarnings("squid:S2189" /* Run forever by design */)
+  @Override
+  public void run() {
+    while (true) {
+      try {
+        doRun();
```
jhunzik: ❓  Should we add some sort of delay here so we don't flood transform with status requests?

peterhuffer: We could but I personally don't mind flooding right now. Eventually I think we will have a back-off time and depending on the amount of failures for a status.

ahoffer: https://github.com/dyng/polling provides strategy objects for for waiting and for giving up:

These are the strategies for waiting between polling attempts:

- FixedWaitStrategy
- RandomWaitStrategy
- CompositeWaitStrategy
- FibonacciWaitStrategy
- ExponentialWaitStrategy
- IncrementingWaitStrategy

These are the strategies for deciding when to give up:

- NeverStopStrategy
- StopAfterAttemptStrategy
- StopAfterDelayStrategy
- CompositeStopStrategy

src/main/java/com/connexta/store/poller/TransformStatusPoller.java
```diff
@@ -0,0 +1,270 @@
...
+        return;
+      }
+    }
+  }
+
+  @VisibleForTesting
+  void doRun() throws InterruptedException {
+    TransformStatusTask transformStatusTask = transformStatusQueue.take();
+    boolean shouldRetry = handleTask(transformStatusTask);
+    if (shouldRetry) {
```
ahoffer: :exclamation: It looks like there are cases where a transform request will never leave the queue. What happens if the endpoint's server has a problem and holds the request connection open forever? What happens if a job gets stuck in the IN_PROGRESS state and never advances?

peterhuffer: It would currently keep putting on the back of the queue, so it would cycle through, but some would get stuck in the queue. The quick solution right now would be to add a last tried timestamp and a failure attempt to the TransformStatusTask and only try a poll if 5 seconds have passed. And if we exceed X failures, just log it and remove the task completely.

ahoffer: That would work. I think just the timestamp would be sufficient. No need to do both.

peterhuffer: I ended up adding back in the dynger polling library to use their retry and stop strategies.

src/main/java/com/connexta/store/poller/TransformStatusPoller.java
```diff
@@ -0,0 +1,270 @@
...
+    final Status transformationStatus = transformationPollResponse.getTransformationStatus();
+    switch (transformationStatus) {
+      case DONE:
+        return addMetadata(datasetId, transformationPollResponse.getMetadataInformations());
+      case FAILED:
+        logFailedTransformation(datasetId, transformationPollResponse.getMetadataInformations());
+        return quarantine(datasetId, transformationPollResponse.getMetadataInformations());
+      case IN_PROGRESS:
+      default:
+        log.debug("Transformation in progress for Dataset {}", datasetId);
```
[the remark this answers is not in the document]
peterhuffer: Threw an illegal argument exception in default case.

src/main/java/com/connexta/store/poller/TransformStatusPoller.java
```diff
@@ -0,0 +1,270 @@
...
+    log.debug("Metadata transformation failed for dataset {}", datasetId);
+    for (MetadataInformation metadataInformation : metadataInformations) {
+      log.debug("Metadata: {}", metadataInformation);
+    }
+  }
+
+  /** @return {@code true} if the add metadata should be retried, {@code false} otherwise */
+  private boolean addMetadata(String datasetId, List<MetadataInformation> metadataInformations) {
+    AddMetadataRequest addMetadataRequest = new AddMetadataRequest();
+    for (MetadataInformation metadataInformation : metadataInformations) {
```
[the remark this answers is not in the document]
peterhuffer: Transform's is MetadataInformation and Store's is MetadataInfo. I made it that way in the API mostly so I wouldn't have to namespace.

jhunzik:
Hero'd locally with a real Transformation service in a docker container. 
I tested:
- Ingesting a file + metacard
- Retrieving the file
- Retrieving the updated metacard
- Retrieving the generated IRM file
- Ingesting a second file + metacard and repeating the test process.

PR #241, opened 2019-11-27
Adds Jacoco

jhunzik:
Adds Jacoco to the test phase like we did here: https://github.com/connexta/ion-search/pull/72

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,6 +22,7 @@ buildscript {
 
 plugins {
     id "com.diffplug.gradle.spotless" version "3.26.0"
+    id "jacoco"
     id "java"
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -84,6 +85,31 @@ spotless {
...
+        csv.enabled false
+        html.destination file("${buildDir}/reports/jacoco")
+    }
+}
+
+jacocoTestCoverageVerification {
+    violationRules {
+        rule {
+            limit {
+                minimum = 0.39
```
jhunzik: The low coverage is due to StoreITests being commented out. @peterhuffer will take care of that in an upcoming PR.

PR #246, opened 2019-12-03
Implement Dataset unstaging

jhunzik:
**[Ready For Review and Hero]**

This PR implements the unstage method in the StoreService. Unstaging the file and old metacard happens during the call to `addMetadata` in the StoreService. The file's status is updated to "stored" and the old file is deleted. The status of the requested Data in the `getData` method is verified to endure that only "stored" and "staged" Data is retrievable.

Hero steps:

1. Start up a local deployment from the `/ion-store/deployments/local`
2. Use the deploy.groovy script under the `scripts` directory to bring up a stack with the Store service and localstack.
3. Send an ingest request and verify the metacard and file status tags are "staged". 
4. Verify that they can be retrieved. (Staged Data is retrievable because Transform needs to access it. In the long term this will be locked down to just a Transformation service).
5. Send a quarantine request with the datasetId from the previous step and verify that the file and metacard are no longer retrievable.
6. Spin up an instance of the Transformation-Api service in a docker container and expose port 9090.
7. Send another ingest request and this time verify in the logs that the file is now "stored", the metacard has been deleted, and IRM has been stored. 
8. Verify that the file and IRM can be retrieved.
9 Attempting to retrieve the metacard should result in a 404.

### DIFF
--- a/src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
+++ b/src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
@@ -82,8 +82,6 @@ public class S3StorageAdaptor implements StorageAdaptor {
     try {
       PutObjectRequest putObjectRequest =
           new PutObjectRequest(bucket, key, inputStream, objectMetadata);
-      putObjectRequest.withTagging(
-          new ObjectTagging(ImmutableList.of(new Tag(STATUS_KEY, StoreStatus.STAGED))));
       final Upload upload = transferManager.upload(putObjectRequest);
       log.info(String.format("Transfer state: %s", upload.getState()));
       upload.waitForCompletion();
@@ -116,15 +114,16 @@ public class S3StorageAdaptor implements StorageAdaptor {
           MediaType.valueOf(objectMetadata.getContentType()),
           inputStream,
           objectMetadata.getUserMetadata());
-    } catch (final Throwable t) {
+    } catch (final Exception e) {
       if (inputStream != null) {
         try {
           inputStream.close();
-        } catch (final IOException e) {
-          log.warn("Unable to close InputStream when retrieving key \"{}\".", key, e);
+        } catch (final IOException ioe) {
+          log.warn("Unable to close InputStream when retrieving key \"{}\".", key, ioe);
         }
       }
-      throw t;
+      throw new RetrieveException(
+          String.format("Unable to retrieve Data for datasetId:{%s}", key), e);
     }
   }
 
@@ -134,7 +133,9 @@ public class S3StorageAdaptor implements StorageAdaptor {
       final Tag updatedTag = new Tag(STATUS_KEY, status);
       final ObjectTagging objectTagging = new ObjectTagging(ImmutableList.of(updatedTag));
       amazonS3.setObjectTagging(new SetObjectTaggingRequest(bucket, key, objectTagging));
+      log.info("Updated Data with datasetId={} in bucket={} with status={}", key, bucket, status);
     } else {
+      log.warn("Dataset for datasetId={} does not exist", key);
       throw new DatasetNotFoundException(key);
     }
   }
@@ -154,6 +155,7 @@ public class S3StorageAdaptor implements StorageAdaptor {
     return statusTag.get().getValue();
   }
 
+  @Override
   public void delete(@NotBlank String key) {
     log.info("Deleting item in bucket \"{}\" with key \"{}\"", bucket, key);
     try {
--- a/src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
+++ b/src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
@@ -227,6 +227,8 @@ public class StoreServiceImpl implements StoreService {
 
   private void verifyStoreStatus(String datasetId, StorageAdaptor adaptor) {
     String status = adaptor.getStatus(datasetId);
+    // Todo: This should also fail for "staged" statuses once staged retrieval can be locked down to
+    // only the Transformation Service
     if (!(StringUtils.equals(status, STORED) || StringUtils.equals(status, STAGED))) {
       log.info("Retrieval failed. Status for datasetId={{}} is not \"stored\".", datasetId);
       throw new RetrieveException(
--- a/src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
+++ b/src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
@@ -6,11 +6,15 @@
  */
 package com.connexta.store.service.impl;
 
+import static com.connexta.store.adaptors.StoreStatus.STAGED;
+import static com.connexta.store.adaptors.StoreStatus.STORED;
+
 import com.connexta.store.adaptors.StorageAdaptor;
 import com.connexta.store.adaptors.StorageAdaptorRetrieveResponse;
 import com.connexta.store.clients.IndexDatasetClient;
 import com.connexta.store.clients.TransformClient;
 import com.connexta.store.controllers.StoreController;
+import com.connexta.store.exceptions.DatasetNotFoundException;
 import com.connexta.store.exceptions.QuarantineException;
 import com.connexta.store.exceptions.RetrieveException;
 import com.connexta.store.exceptions.StoreException;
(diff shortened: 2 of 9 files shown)


src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
```diff
@@ -73,6 +77,7 @@ public void ingest(
 
     fileStorageAdaptor.store(
         fileSize, mimeType, inputStream, datasetId, Map.of(FILE_NAME_METADATA_KEY, fileName));
+    fileStorageAdaptor.updateStatus(datasetId, STAGED);
```
figliold: ❓ Do storing and updating status need to happen as one unit of work?  In other words, would things break of the store happened but the status was never updated to staged?

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
```diff
@@ -216,4 +216,39 @@ public void quarantine(
...
+      throws DatasetNotFoundException, QuarantineException {
+    fileStorageAdaptor.updateStatus(datasetId, STORED);
+    metacardStorageAdaptor.delete(datasetId);
+    log.info("Unstaged file and deleted staged metacard for datasetId={{}}", datasetId);
+  }
+
+  private void verifyStoreStatus(String datasetId, StorageAdaptor adaptor) {
+    String status = adaptor.getStatus(datasetId);
+    // TODO: Should the exception that is returned depend on what the status is (Staged vs
+    // Quarantined)?
```
figliold: ❓ Should you use !StringUtils.equalsIgnoreCase(status, STORED) instead or an enum just to be safe?

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
```diff
@@ -216,4 +216,39 @@ public void quarantine(
...
+    if (status == null || !status.equals(STORED)) {
+      log.info("Retrieval failed. Status for datasetId={{}} is not \"stored\".", datasetId);
+      throw new RetrieveException(
+          String.format("Dataset for Id={%s} has not been stored", datasetId));
+    }
+  }
+
+  @NotNull
+  private Resource validateMetadataResource(ResponseEntity<Resource> metadataResource) {
+    if (metadataResource == null) {
```
figliold: ❓ I'm not sure if/how we want to handle this.  Should the Service know anything about HTTP status codes?  What if the Service was used outside of an endpoint?  Wouldn't it be weird to use HTTP status codes in that situation?  Thoughts?

jhunzik: This was just refactored from `addMetadata` to reduce complexity. @peterhuffer could chime in. My two cents are to throw a custom exception.

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
jhunzik: Most people have suggested an enum, so I'll go ahead and change that.

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
peterhuffer: @figliold no we shouldn't be. These interfaces should define exceptions that are thrown and caught by an exception handler which will translate them to the appropriate HTTP response. Our interfaces were throwing these HTTP response exceptions though already so this is something we need to address later.

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
jhunzik: If that happened, you still wouldn't be able to retrieve the file or metacard. We only retrieve if the status is "stored". I do think your comment applies to the whole ingest flow. Ie it breaks after storing the file, but before storing the metacard. I'm not sure how we account for that.

peterhuffer: Typically we have been taking the approach of retrying. We shouldn't break, we would just be in a state where the dataset is not retrievable.

peterhuffer: Typically we have been relying on retries. Unfortunately yes, with these methods we could end up in a state where the file is retrievable, but the metacard is not, which is not desirable. I think the goal is that either all data in a dataset is retrievable, or none of it is. Unfortunately right now we have multiple Data sharing the same ID (the datasetId), which is abusing the DatasetId. Each data should have its own ID and a reference to the datasetId which tells it how it is related to other Data.

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
jhunzik: I'm going through and updating to an enum, but I just realized that doesn't fix the check here. The storage adaptor shouldn't take or return an enum. We're stuck with string comparisons for this one.

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
```diff
@@ -216,4 +216,39 @@ public void quarantine(
       adaptor.delete(datasetId);
     }
   }
+
+  @Override
+  public void unstage(
+      @Pattern(regexp = "^[0-9a-zA-Z]+$") @Size(min = 36, max = 36) final String datasetId)
```
peterhuffer: ✏️ This pattern regex is incorrect for a 36 length UUID. These annotations don't even do anything here. I'd rather just not have it.

src/test/java/com/connexta/store/adaptors/S3StorageAdaptorTests.java
```diff
@@ -0,0 +1,130 @@
...
+    // given
+    S3Object mockS3Object = mock(S3Object.class);
+    InputStream mockInputStream = mock(InputStream.class);
+    when(mockAmazonS3.doesObjectExist(eq(BUCKET), anyString())).thenReturn(true);
+    when(mockAmazonS3.getObject(any(GetObjectRequest.class))).thenReturn(mockS3Object);
+    when(mockAmazonS3.doesBucketExistV2(eq(BUCKET))).thenReturn(true);
+    when(mockS3Object.getObjectContent()).thenThrow(new SdkClientException("test"));
+
+    // verify
+    assertThrows(SdkClientException.class, () -> storageAdaptor.retrieve(DATASET_ID));
```
peterhuffer: ❓ What HTTP status code does an SdkClientException translate into? Can we throw the RetrieveException instead of letting this exception leak out of the StorageAdapter interface?

jhunzik: I think it's just 500 from the client's perspective. I'll fix it

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
```diff
@@ -116,15 +114,16 @@ public StorageAdaptorRetrieveResponse retrieve(@NotBlank final String key)
           MediaType.valueOf(objectMetadata.getContentType()),
           inputStream,
           objectMetadata.getUserMetadata());
-    } catch (final Throwable t) {
+    } catch (final Exception e) {
```
jhunzik: I could be wrong, but if anything more severe than Exception gets thrown, the service is already in a non recoverable state. At that point it doesn't matter if the stream is closed or not.

peterhuffer: Throwable will catch all RuntimeExceptions as well if that was the intention here.

src/test/java/com/connexta/store/adaptors/S3StorageAdaptorITests.java
```diff
@@ -211,16 +210,6 @@ void testDeleteBucketDoesNotExist() {
     amazonS3.createBucket(BUCKET);
   }
 
-  @Test
```
jhunzik: This just got moved to the new `S3StorageAdaptorTests` unit test class

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
jhunzik: Ah yeah, that's a good point. I think that was the intention. I'll revert it.

peterhuffer: Maybe just catch Exception and RuntimeException separately? We really shouldn't be catching Throwable.

PR #247, opened 2019-12-03
Fix and re-enable some test

peterhuffer:
[Ready for Review]

- Reenables some of the main thread testing for the StoreITests
- Fixes a bug where StoreController was using the storeApiVersion for the ingestApiVersion
- Fixes quarantine endpoint returning 202 rather than 200


src/test/java/com/connexta/store/StoreITests.java
```diff
@@ -6,541 +6,388 @@
...
+          .isOk()
+          .expectHeader()
+          .contentType(expectedContentType)
+          .expectHeader()
+          .contentDisposition(
+              ContentDisposition.builder("attachment").filename(expectedFilename).build())
+          .expectBody(String.class)
+          .isEqualTo(expectedContent);
+    } catch (AssertionError e) {
+      return false;
```
peterhuffer: Idk how I feel about catching this assertion error exception, but it was the easiest way to deal with awaitility and still use assertions on the WebTestClient

src/test/java/com/connexta/store/StoreITests.java
```diff
@@ -6,541 +6,388 @@
...
-  //
-  //  @Test
-  //  public void testContextLoads() {}
-  //
-  //  /**
-  //   * @see #testRetrieveFileWhenS3IsEmpty()
-  //   * @see StoreControllerRetrieveFileComponentTest#testS3KeyDoesNotExist()
-  //   */
-  //  @Test
-  //  public void testRetrieveFileNotFound() throws Exception {
```
peterhuffer: These tests that tested retrieving specific content I felt were not longer needed because we only deal in data types now.

src/test/java/com/connexta/store/StoreITests.java
```diff
@@ -6,541 +6,388 @@
...
+    TEST_INGEST_REQUEST_BODY = builder.build();
+  }
+
+  private MockWebServer mockTransformServer;
+  private MockWebServer mockFileServer;
+
+  @BeforeEach
+  void beforeEach() throws Exception {
+    mockTransformServer = new MockWebServer();
+    mockTransformServer.start(9090);
```
peterhuffer: ✏️ This had to match the port of the transform service passed in via the test spring config in order for the mocker server to properly return the mock responses. I couldn't figure out how to start the mock server on a random port, then convey that to spring as the transform service's port. Any ideas?

src/main/java/com/connexta/store/poller/StatusService.java
```diff
@@ -64,6 +64,7 @@ public void init() {
               submit(transformStatusQueue.take());
             } catch (InterruptedException e) {
               Thread.currentThread().interrupt();
+              break;
```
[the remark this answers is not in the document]
peterhuffer: I wasn't aware of the Async annotation. Would probably be better to use. I think we are handling them correctly though. Currently we do not care about the task being executed, but ultimately we will.

src/main/java/com/connexta/store/adaptors/S3StorageAdaptor.java
```diff
@@ -160,7 +160,7 @@ public void delete(@NotBlank String key) {
     log.info("Deleting item in bucket \"{}\" with key \"{}\"", bucket, key);
     try {
       if (!s3ObjectExists(bucket, key)) {
```
jhunzik:
```suggestion
      if (s3ObjectExists(bucket, key)) {
          amazonS3.deleteObject(bucket, key);
      }
```

PR #252, opened 2019-12-05
Update index-api version to 0.6.0

peterhuffer:
[Ready for Review]

Depends on https://github.com/connexta/ion-store/pull/247


src/main/java/com/connexta/store/clients/IndexClientImpl.java
```diff
@@ -0,0 +1,61 @@
...
+    final IndexRequest indexRequest =
+        new IndexRequest()
+            .fileLocation(fileUrl.toString())
+            .irmLocation(irmUrl.toString())
+            .metacardLocation(metacardUrl.toString());
+
+    ClientResponse response =
+        indexWebClient
+            .put()
+            .uri(UriComponentsBuilder.fromPath("/index/{datasetId}").build(datasetId.toString()))
```
emmberk: ✏️ Seems like there should be a way to add `"/index/{datasetId}"` to the WebClient builder in `IndexDatasetClientConfiguration` and only use the `uriVariables` here.

emmberk: ✏️ Does `uri("/index/{datasetId}", datasetId.toString())` work instead of using `UriComponentsBuilder`?

src/main/java/com/connexta/store/clients/IndexClientImpl.java
```diff
@@ -0,0 +1,64 @@
...
+    ClientResponse response =
+        indexWebClient
+            .put()
+            .uri(
+                UriComponentsBuilder.fromPath("/index/{datasetId}")
+                    .build(datasetId)
+                    .toASCIIString())
+            .body(BodyInserters.fromValue(indexRequest))
+            .exchange()
+            .block();
```
emmberk: ✏️ Should we add a timeout here? `block(Duration)`

src/main/java/com/connexta/store/clients/IndexClientImpl.java
```diff
@@ -0,0 +1,64 @@
...
+            .irmLocation(irmUrl.toString())
+            .metacardLocation(metacardUrl.toString());
+
+    ClientResponse response =
+        indexWebClient
+            .put()
+            .uri(
+                UriComponentsBuilder.fromPath("/index/{datasetId}")
+                    .build(datasetId)
+                    .toASCIIString())
```
emmberk: ✏️ Seems like there should be a way to set this url template in the WebClient builder so that you only need the uriVariables here.

emmberk: ✏️ Can this be made simpler? `.uri("/index/{datasetId}", datasetId.toString())`

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
```diff
@@ -195,23 +187,33 @@ public void addMetadata(String datasetId, List<MetadataInfo> metadataInfos) thro
...
-                  .path(StoreController.RETRIEVE_DATA_URL_TEMPLATE)
-                  .build(datasetId, IRM_TYPE));
           break;
         default:
           throw new IllegalArgumentException(
               String.format("Received unsupported dataType {%s}", dataType));
       }
     }
+
+    unstage(datasetId);
```
emmberk: ❓ Why is `unstage` in the StoreService interface if it is inly used in `StoreServiceImpl`?

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
```diff
@@ -195,23 +187,33 @@ public void addMetadata(String datasetId, List<MetadataInfo> metadataInfos) thro
...
+
+    indexClient.indexDataset(
+        UUID.fromString(datasetId),
```
emmberk: ✏️ I prefer keeping the datasetId as a UUID always until we need it as a String so we don't need to validate it everywhere besides checking non-null. I suggest updating he method param to be a UUID.

src/test/java/com/connexta/store/StoreITests.java
```diff
@@ -181,7 +178,8 @@ void beforeEach() throws Exception {
     mockFileServer = new MockWebServer();
     mockFileServer.start();
 
-    indexMockRestServiceServer = MockRestServiceServer.createServer(nonBufferingRestTemplate);
+    mockIndexServer = new MockWebServer();
+    mockIndexServer.start();
```
emmberk: ❓ Why is `mockIndexServer` created? I don't see it being used anywhere?

src/test/java/com/connexta/store/StoreITests.java
```diff
@@ -127,13 +125,11 @@ public AmazonS3Configuration testAmazonS3Configuration() {
...
 
-  private MockRestServiceServer indexMockRestServiceServer;
-
   @Inject private AmazonS3 amazonS3;
 
   @Inject
-  @Named("nonBufferingRestTemplate")
-  private RestTemplate nonBufferingRestTemplate;
+  @Named(IndexClientConfiguration.INDEX_WEB_CLIENT_BEAN)
+  private WebClient indexWebClient;
```
emmberk: ❓ Why was this injected? I don't see it being used anywhere?

src/test/java/com/connexta/store/StoreITests.java
```diff
@@ -223,10 +219,6 @@ void testRetrieveDataNotFound() {
   @Test
   void testIngestRequest() throws Exception {
```
emmberk: ❓ I don't see a test for `addMetadata` in this class, but I'd expect there to be one here where we stub the indexServer.

src/test/java/com/connexta/store/WebClientTestUtils.java
```diff
@@ -0,0 +1,72 @@
...
+public class WebClientTestUtils {
+
+  private WebClientTestUtils() {}
+
+  /**
+   * Works by passing in a mock {@link ExchangeFunction}, which is used to capture the {@link ClientRequest}
+   * that is sent through the {@link WebClient}. For example,
+   *
+   * <pre>
+   * {@code
```
emmberk: ✏️ Looks like this is missing the closing bracket.

src/test/java/com/connexta/store/clients/IndexClientImplTest.java
```diff
@@ -0,0 +1,98 @@
...
+    try {
+      FILE_URL = new URL("http://localhost:9090/file");
+      IRM_URL = new URL("http://localhost:9090/irm");
+      METACARD_URL = new URL("http://localhost:9090/metacard");
+    } catch (MalformedURLException e) {
+      throw new RuntimeException();
+    }
+  }
+
+  private IndexClient indexClient;
```
emmberk: ✏️ I don't see this being used anywhere

src/test/java/com/connexta/store/clients/IndexClientImplTest.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  @ParameterizedTest
+  @ValueSource(ints = {400, 401, 403, 500, 501})
+  void testNonSuccessHttpResponse(int statusCode) {
+    // setup
+    ExchangeFunction exchangeFunction = mock(ExchangeFunction.class);
+    WebClient webClient =
+        WebClientTestUtils.mockEmptyResponseWebClient(
+            HttpStatus.valueOf(statusCode), exchangeFunction);
```
emmberk: ✏️ 
```java
@ParameterizedTest
@EnumSource(
    value = HttpStatus.class,
    names = {"OK"},
    mode = EnumSource.Mode.EXCLUDE)
void testNonSuccessHttpResponse(HttpStatus httpStatus) {
  // setup
  ExchangeFunction exchangeFunction = mock(ExchangeFunction.class);
  WebClient webClient =
      WebClientTestUtils.mockEmptyResponseWebClient(
          httpStatus, exchangeFunction);
```

src/test/java/com/connexta/store/clients/IndexClientImplTest.java
```diff
@@ -0,0 +1,98 @@
...
+    // when
+    indexClient.indexDataset(DATASET_ID, FILE_URL, IRM_URL, METACARD_URL);
+
+    // then
+    ArgumentCaptor<ClientRequest> indexRequestCaptor = ArgumentCaptor.forClass(ClientRequest.class);
+    verify(exchangeFunction).exchange(indexRequestCaptor.capture());
+
+    ClientRequest indexRequest = indexRequestCaptor.getValue();
+    assertThat(indexRequest.method(), is(HttpMethod.PUT));
+    assertThat(indexRequest.url().toASCIIString(), is("/index/" + DATASET_ID));
```
emmberk: ✏️ Use static string for the index url template in the tests and client

src/test/java/com/connexta/store/service/impl/StoreServiceImplTest.java
```diff
@@ -178,31 +180,43 @@ void testSuccessfulIngest(
...
             transformClient,
             taskQueue,
             mockTransformWebClient);
-    var metadataInfo = new MetadataInfo();
-    metadataInfo.setLocation(new URL("http://location"));
-    metadataInfo.setMetadataType("irm");
+    var irmMetadataInfo = new MetadataInfo();
+    irmMetadataInfo.setLocation(new URL("http://location:1234/irm"));
+    irmMetadataInfo.setMetadataType("irm");
+
```
emmberk: ✏️ Can instantiate in one line
```java
new MetadataInfo().location(new URL("http://location:1234/metacard")).metadataType("metacard");
```

src/test/java/com/connexta/store/service/impl/StoreServiceImplTest.java
```diff
@@ -338,15 +325,15 @@ void testGettingDatasetNotStored() {
...
+    verifyNoInteractions(indexClient);
   }
 
   @Test
   void testGettingDatasetNullStatus() {
     when(metacardStorageAdaptor.getStatus(eq(DATASET_ID))).thenReturn(null);
 
     assertThrows(RetrieveException.class, () -> storeService.getData(DATASET_ID, "metacard"));
-    verifyNoInteractions(indexDatasetClient);
+    verifyNoInteractions(indexClient);
```
emmberk: ✏️ Move `verifyNoInteractions(indexClient)` to `@AfterEach` because this should be verified for every test method

src/main/java/com/connexta/store/clients/IndexClient.java
```diff
@@ -0,0 +1,15 @@
...
+ * https://www.gnu.org/licenses/lgpl-3.0.html
+ */
+package com.connexta.store.clients;
+
+import java.net.URL;
+import java.util.UUID;
+
+public interface IndexClient {
+
+  void indexDataset(UUID datasetId, URL fileUrl, URL irmUrl, URL metacardUrl);
```
emmberk: ✏️ Document exceptions

src/main/java/com/connexta/store/clients/IndexClientImpl.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.web.reactive.function.BodyInserters;
+import org.springframework.web.reactive.function.client.ClientResponse;
+import org.springframework.web.reactive.function.client.WebClient;
+import org.springframework.web.util.UriComponentsBuilder;
+
+@Slf4j
+@AllArgsConstructor
+public class IndexClientImpl implements IndexClient {
+
+  public static final String ACCEPT_VERSION_HEADER_NAME = "Accept-Version";
```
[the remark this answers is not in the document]
peterhuffer: Yup its not being used. Removed.

src/main/java/com/connexta/store/clients/IndexClientImpl.java
peterhuffer: I could, but then the client would no longer be specific to the host. Ie instead of `http://store:8080`, it would be `http://store:8080/index`. If someone wanted to do an options for example, they would then have to make their own client.

src/test/java/com/connexta/store/StoreITests.java
peterhuffer: I removed it since the flow is ingest should be called. Since we expect a file to always be present before calling addMetadata, there are issues with calling this method before ingest.

src/main/java/com/connexta/store/clients/IndexClientImpl.java
peterhuffer: I haven't been, but yeah that would probably be smart. Would it be okay to address this in a separate PR in a global way? Also we may want to make it configurable.

src/test/java/com/connexta/store/StoreITests.java
emmberk: So there are no itests for addMetadata? quarantine has this same problem. I think we need a test for addMetadata to make sure that everything is wired up correctly.

src/main/java/com/connexta/store/service/impl/StoreServiceImpl.java
peterhuffer: Not sure. I could potentially see it being used later from the controller. We should leave it as is until we re-do these interfaces.

src/main/java/com/connexta/store/clients/IndexClientImpl.java
emmberk: Yeah. A separate PR sounds good.

src/test/java/com/connexta/store/StoreITests.java
peterhuffer: This may not be needed.

src/test/java/com/connexta/store/StoreITests.java
peterhuffer: Whoops, I should be returning a successful request from here. Not sure why the test worked! I wonder if default is success.

src/test/java/com/connexta/store/clients/IndexClientImplTest.java
peterhuffer: I only want to handle what the API says it will return. Will this include every reponse code from the HttpStatus class excluding OK?

src/test/java/com/connexta/store/StoreITests.java
peterhuffer: No. AddMetadata is exercised through Ingest. Ingest tests cannot pass without addMetadata working.

peterhuffer: Ingest is the primary work flow and entry point into CDR. In fact I should probably probably modify the quarantine test to have transform return a response to trigger it rather than call quarantine directly.

src/test/java/com/connexta/store/clients/IndexClientImplTest.java
emmberk: Yes, that's how it would work. I'm fine only testing the exact status codes defined in the API.

src/test/java/com/connexta/store/StoreITests.java
emmberk: Oh, I see that you stub the transform status server in ingest so that the poller will call addMetadata 👍.

src/test/java/com/connexta/store/service/impl/StoreServiceImplTest.java
peterhuffer: This guy doesn't have a builder implementation.

src/main/java/com/connexta/store/clients/IndexClient.java
peterhuffer: I opted to just remove and not document. Since they are runtime exceptions which are translated into HTTP responses, they should never be caught by a caller anyways. Eventually we need to add checked exceptions to this.

src/test/java/com/connexta/store/service/impl/StoreServiceImplTest.java
emmberk: https://github.com/connexta/ion-store-api/blob/2a881be13ab0b57f39789562c17c1d5955b44343/rest/models/src/main/java/com/connexta/store/rest/models/MetadataInfo.java#L19-L22
https://github.com/connexta/ion-store-api/blob/2a881be13ab0b57f39789562c17c1d5955b44343/rest/models/src/main/java/com/connexta/store/rest/models/MetadataInfo.java#L35-L38

src/test/java/com/connexta/store/clients/IndexClientImplTest.java
peterhuffer: I personally prefer the test and impl have their own separate ones so that the tests break if someone mistakenly changes it.

PR #254, opened 2019-12-06
Fix local deployment readme

emmberk:


deployments/local/README.md
```diff
@@ -12,17 +12,16 @@ and irm. A basic dashboard UI is available on port 8055 and a JSON record of all
 * Linux/Mac
```
jhunzik: ✏️ The last sentence of the description can be removed. That directory gets created automatically when Localstack starts up.

